PR #2, opened 2022-09-27
Add support for capturing BOSH VMs

peanball:
Add logic to capture from BOSH VMs

Authorization via BOSH UAA token

Optional allow-list on pcap-api to limit CIDRs from which requests are accepted (via `X-Forwarded-For` header)


maxmoehl:
Why did you change the go module structure? I know that almost all other releases do it but I don't like that it forces you to always open the whole project compared to being able to work on a single part of it.

peanball:
Closed in favour of #5 from the sap-contributions repo

PR #17, opened 2023-03-14
Implement bosh integration

a18e:


src/pcap/cmd/mock-bosh-api/main.go
```diff
@@ -0,0 +1,158 @@
+package main
```
peanball, line 1: Consider moving this file into the test package, as it is not to be used for the actual release.

src/pcap/cmd/pcap-api/config.go
```diff
@@ -26,19 +27,18 @@ var DefaultAPIConfig = APIConfig{
...
+	DrainTimeout:       10 * time.Second,
+	ConcurrentCaptures: 5,
 }
 
 type APIConfig struct {
-	cmd.CommonConfig
-	Agents             *pcap.AgentMTLS `yaml:"agents"`
-	ConcurrentCaptures int             `yaml:"concurrent_captures"`
-	DrainTimeout       time.Duration   `yaml:"drain_timeout"`
+	cmd.CommonConfig   `yaml:"common_config"` // fixme: workaround by a18e to be able to parse a config file again, may not be desired
```
peanball, line 35: I don't like this whole contraption. I will work a bit on resolving this "common" `cmd` package and, the `CommonConfig` etc.

src/pcap/cmd/pcap-api/main.go
```diff
@@ -48,7 +48,18 @@ func main() {
 		log.Fatal("unable to create api", zap.Error(err))
 	}
 
-	api.RegisterResolver(&pcap.BoshHandler{Config: config.ManualEndpoints})
+	// set up a BoshResolver for each bosh environment
+	for _, env := range config.BoshEnvironments {
+		resolver, err := pcap.NewBoshResolver(env, 8083) // FIXME
```
peanball, line 53: The port should be part of the BOSH configuration, which contains a list of `Environment`s.

src/pcap/cmd/pcap-bosh-cli/main.go
```diff
@@ -0,0 +1,285 @@
...
+	setupLogging()
+
+	defer func() {
+		if err != nil {
+			logger.Panic("execution failed", zap.Error(err))
+		}
+	}()
+
+	// Parse command-line arguments
+	_, err = flags.ParseArgs(&opts, os.Args[1:])
```
peanball, line 56: Because we're not doing anything special with the args before parsing, we can also just use the default function `Parse()` here. It automatically takes `os.Args`.
```suggestion
	_, err = flags.Parse(&opts)
```

src/pcap/cmd/pcap-bosh-cli/main.go
```diff
@@ -0,0 +1,285 @@
...
+			logger.Info("capture finished successfully")
+			return
+		}
+		err = fmt.Errorf("finished with error %w", err)
+		return
+	}
+}
+
+// sets up the zap.Logger. Currently outputs to stderr in Console format.
+func setupLogging() {
```
peanball, line 131: this could also be `init()` after all, like in the other executables.

src/pcap/cmd/pcap-bosh-cli/main.go
```diff
@@ -0,0 +1,285 @@
...
+	atomicLogLevel = zap.NewAtomicLevelAt(zap.WarnLevel)
```
peanball, line 132: It's a CLI, so `InfoLevel` would probably be alright as default?

src/pcap/cmd/pcap-bosh-cli/main.go
```diff
@@ -0,0 +1,285 @@
...
+	case 1:
+		logLevel = zapcore.InfoLevel
+	default: // if more than one -v is given as argument
+		logLevel = zapcore.DebugLevel
+	}
+
+	atomicLogLevel.SetLevel(logLevel)
+
+	logger.Debug("set log-level", zap.String("log-level", logLevel.String()))
+}
```
peanball, line 168: This is taken from the `pcap-api`. Because this is a CLI, default could / should be INFO.

src/pcap/cmd/pcap-bosh-cli/main.go
```diff
@@ -0,0 +1,285 @@
...
+	}
+	if parsedURL.Scheme != "https" && parsedURL.Scheme != "http" {
+		return nil, fmt.Errorf("invalid URL, must start with http or https: %s", parsedURL)
+	}
+	return parsedURL, nil
+}
+
+// getEnvironment searches the provided bosh.Config (config) for the environmentAlias.
+//
+// Returns a matching environment if one exists.
```
peanball, line 216:
```suggestion
// Returns a matching environment if one exists, or an error if it doesn't.
```

peanball (commented):
First part of the review. Next is the package `pcap`.

src/pcap/config/config.example.yml
```diff
@@ -44,13 +44,24 @@ api:
       private_key: cc-client-key.key
       ca: cc-ca.pem
   bosh: # -> *pcap.BoshHandler `yaml:omitempty`
```
peanball, line 46: consider extending `Environment` with the `AgentPort` or adding a property here. Probably it's easier and more flexible to add the AgentPort to `Environment`. We can also make a new composite struct that contains an `Environment` and additional things.

We can then write out the pure `Environment` bit of it when writing out the `~/.bosh/config` file in the pcap-bosh-cli.

src/pcap/test/mocks.go
```diff
@@ -0,0 +1,320 @@
...
+
+func MustParseURL(rawURL string) *url.URL {
+	url, err := url.Parse(rawURL)
+	if err != nil {
+		panic(err)
+	}
+	return url
+}
+
+func MockJwtAPI() (*httptest.Server, string) {
```
peanball, line 31: If you wanted to keep the Mock Bosh Director around as executable, it should share code with this.

src/pcap/agent.go
```diff
@@ -234,13 +234,13 @@ func readPackets(ctx context.Context, cancel CancelCauseFunc, handle pcapHandle,
 				return
 			}
 
-			data, _, err := handle.ReadPacketData()
+			data, ci, err := handle.ReadPacketData()
```
peanball, line 237:
```suggestion
			data, captureInfo, err := handle.ReadPacketData()
```

src/pcap/agent.go
```diff
@@ -234,13 +234,13 @@ func readPackets(ctx context.Context, cancel CancelCauseFunc, handle pcapHandle,
...
 			if err != nil {
 				cancel(fmt.Errorf("read packet: %w", err))
 				return
 			}
 
-			out <- newPacketResponse(data)
+			out <- newPacketResponse(data, ci)
```
peanball, line 243:
```suggestion
			out <- newPacketResponse(data, captureInfo)
```

src/pcap/bosh_resolver.go
```diff
@@ -0,0 +1,208 @@
...
+
+type BoshResolver struct {
+	environment bosh.Environment
+	client      *http.Client
+	uaaURLs     []string
+	agentPort   int
+}
+
+func NewBoshResolver(environment bosh.Environment, agentPort int) (*BoshResolver, error) {
+	resolver := &BoshResolver{environment: environment, agentPort: agentPort} // TODO: get agent port from where?
```
peanball, line 28: I think adding the port to an extended `Environment` that augments the regular BOSH config "Environment" might be the way to go.

src/pcap/bosh_resolver.go
```diff
@@ -0,0 +1,208 @@
...
+	err := resolver.setup()
+	if err != nil {
+		return nil, err
+	}
+	return resolver, err
+}
+
+func (br *BoshResolver) Name() string {
+	// TODO we need to differentiate between bosh resolvers for different environments (i.e. bootstrap-bosh, bosh. This would take some refactoring, e.g. in pcap.proto StatusResponse
```
peanball, line 37: So do you want to signal, which environments are supported on the server as part of the status message?

We could consider having a `resolvers` field that just contains the names (and types?) of all known resolvers, instead of having the hardcoded `cf` and `bosh` boolean flags.

src/pcap/bosh_resolver.go
```diff
@@ -0,0 +1,208 @@
...
+	// return fmt.Sprintf("bosh/%s", br.environment.Alias)
+	return "bosh"
+}
+
+func (br *BoshResolver) CanResolve(request *EndpointRequest) bool {
+	return request.GetBosh() != nil
```
peanball, line 43: Here we should then also check that this is for our environment.

src/pcap/bosh_resolver.go
```diff
@@ -0,0 +1,208 @@
...
+	boshRequest := request.GetBosh()
+
+	err = br.authenticate(boshRequest.Token)
+	if err != nil {
+		return nil, err
+	}
+
+	instances, _, err := br.getInstances(boshRequest.Deployment, boshRequest.Token)
+	if err != nil {
+		log.Error("failed to get instances from bosh-director", zap.String(LogKeyTarget, ""))
```
peanball, line 64: Can we have a value for `LogKeyTarget`?

src/pcap/bosh_resolver.go
```diff
@@ -0,0 +1,208 @@
...
+		return nil, err
+	}
+
+	var endpoints []AgentEndpoint
+	for _, instance := range instances {
+		identifier := strings.Join([]string{instance.Job, instance.Id}, "/")
+		endpoints = append(endpoints, AgentEndpoint{IP: instance.Ips[0], Port: br.agentPort, Identifier: identifier}) //TODO: defaultport ok here?
```
peanball, line 71: yes, default port is OK here.

src/pcap/bosh_resolver.go
```diff
@@ -0,0 +1,208 @@
...
+	}
+
+	if len(endpoints) == 0 {
+		return nil, fmt.Errorf("no matching endpoints found")
```
peanball, line 75: Declare this particular error as variable so it could be reused and unwrapped, if needed.

src/pcap/bosh_resolver.go
```diff
@@ -0,0 +1,208 @@
...
+	}
+
+	log.Debug("received AgentEndpoints from Bosh Director", zap.Any("agent-endpoint", endpoints))
+	return endpoints, nil
+}
+
+func (br *BoshResolver) validate(endpointRequest *EndpointRequest) error {
```
peanball, line 82: validate needs adaptation to the new semantics (e.g. the `Environment` needs to be set)

src/pcap/bosh_resolver.go
```diff
@@ -0,0 +1,208 @@
...
+
+func (br *BoshResolver) setup() error {
+	log.Infof("Setting Up BoshResolver for %s", br.environment.Alias)
+
+	if br.environment.CaCert == "" {
+		br.client = http.DefaultClient
+	} else {
+		data, err := os.ReadFile(br.environment.CaCert)
+		if err != nil {
+			return fmt.Errorf("could not load BOSH Director CA from %s (%s)", br.environment.CaCert, err)
```
peanball, line 112:
```suggestion
			return fmt.Errorf("could not load BOSH Director CA from %q: %w", br.environment.CaCert, err)
```
Wrapping errors with `%w` allows to unwrap them later and to check for their existence down the chain of wrapped erros with `errors.Is(err, someErrorThatCouldBeWrappedHere)`.

src/pcap/bosh_resolver.go
```diff
@@ -0,0 +1,208 @@
...
+		return fmt.Errorf("could not read BOSH Director API response: %s", err)
+	}
+	err = json.Unmarshal(data, &apiResponse)
+	if err != nil {
+		return fmt.Errorf("could not parse BOSH Director API response: %s", err)
+	}
+
+	br.uaaURLs = apiResponse.UserAuthentication.Options.Urls
+
+	log.Infof("Connected to BOSH Director '%s' (%s), version %s on %s. UAA URLs: %v", apiResponse.Name, apiResponse.Uuid, apiResponse.Version, apiResponse.Cpi, br.uaaURLs)
```
peanball, line 150: this is still using logrus.

src/pcap/bosh_resolver.go
```diff
@@ -0,0 +1,208 @@
...
+
+	return nil
+}
+
+func (br *BoshResolver) authenticate(authToken string) error {
+
+	allowed, err := VerifyJwt(authToken, "bosh.admin", br.uaaURLs)
+	if err != nil {
+		return fmt.Errorf("could not verify token %s (%s)", authToken, err)
```
peanball, line 159: add the error after `:`, wrap it.

src/pcap/bosh_resolver.go
```diff
@@ -0,0 +1,208 @@
...
+	data, err := io.ReadAll(res.Body)
+	if err != nil {
+		return nil, 0, err
+	}
+
+	if res.StatusCode != http.StatusOK {
+		return nil, res.StatusCode, fmt.Errorf("expected status code %d but got status code %d: %s", http.StatusOK, res.StatusCode, string(data))
+	}
+
+	var response []bosh.Instance
```
peanball, line 200: not specific to this line, but having a package `bosh` for some of the types is a bit weird. Either put the whole resolver into a package `bash` or all code into the `bash.go` in package `pcap`.

peanball (commented):
next batch of files reviewed.

continuing at bosh_resolver_test.go

src/pcap/client.go
```diff
@@ -0,0 +1,240 @@
...
+		CaptureLength:  len(packet.Data),
+		Length:         int(packet.Length),
+		InterfaceIndex: 0,
+		AncillaryData:  nil,
+	}
+	err := packetWriter.WritePacket(captureInfo, packet.Data)
+	if err != nil {
+		zap.L().Error("writing packet to file failed", zap.Error(err))
+	}
+	zap.L().Info("received packet", zap.Int("bytes", len(packet.Data)), zap.Time("capture-timestamp", packet.Timestamp.AsTime()))
```
peanball, line 208: You can't log every received packet. And definitely not at Info level.

src/pcap/client.go
```diff
@@ -0,0 +1,240 @@
...
+}
+
+// TODO: still needed?
```
peanball, line 211: yes, still needed. you can't log every single packet.

src/pcap/client.go
```diff
@@ -0,0 +1,240 @@
...
+func writeMessage(message *Message) {
+	var logLevel zapcore.Level
+
+	switch message.Type {
+	case MessageType_UNKNOWN:
+		logLevel = zapcore.WarnLevel
+	case MessageType_INSTANCE_UNAVAILABLE:
+		logLevel = zapcore.WarnLevel
+	case MessageType_START_CAPTURE_FAILED:
+		logLevel = zapcore.WarnLevel
```
peanball, line 180: error? When you can't start the capture at all...

src/pcap/client.go
```diff
@@ -0,0 +1,240 @@
...
+	case MessageType_CONGESTED:
+		logLevel = zapcore.WarnLevel
+	case MessageType_LIMIT_REACHED:
+		logLevel = zapcore.WarnLevel
+	case MessageType_CAPTURE_STOPPED:
+		logLevel = zapcore.InfoLevel
+	case MessageType_CONNECTION_ERROR:
+		logLevel = zapcore.ErrorLevel
+	}
+	log.Infof("%s - %v", message.String(), logLevel.String()) // TODO: duplicate logging
```
peanball, line 192: this is logrus. Should not be there.

src/pcap/client_test.go
```diff
@@ -0,0 +1,287 @@
...
+	_ = os.Remove(outputFile)
+
+	tests := []struct {
+		name              string
+		outputFile        string
+		wantErr           bool
+		expectedErrString string
+	}{
+		{
+			name:       "first run - file does not exist",
```
peanball, line 165:
```suggestion
			name:       "output file does not exist",
```

1. there should not be a deterministic order for test cases.
2. The label 'first run' does not add any information.

src/pcap/client_test.go
```diff
@@ -0,0 +1,287 @@
...
+			outputFile: outputFile,
+			wantErr:    false,
+		}, {
+			name:              "second run - file does exist",
```
peanball, line 169:
```suggestion
			name:              "output file does exist",
```

src/pcap/client_test.go
```diff
@@ -0,0 +1,287 @@
...
+//			}
+//			if tt.expectedErr != nil && !errors.Is(err, tt.expectedErr) {
+//				t.Errorf("expectedErr = %v, actualErr = %v", tt.expectedErr, err)
+//			}
+//			if client == nil {
+//				t.Errorf("client is nil")
+//			}
+//		})
+//	}
+//}
```
peanball, line 232: yes. Move to the integration test.

src/pcap/client_test.go
```diff
@@ -0,0 +1,287 @@
...
+
+// writes a predefined packet to a pcap-file, parses the file and compares timestamp and destination port
```
peanball, line 234:
```suggestion
// TestWritePacket writes a predefined packet to a pcap-file, parses the file and compares timestamp and destination port
```

src/pcap/client.go
```diff
@@ -0,0 +1,240 @@
...
+			return
+		}
+	}
+}
+
+func tryCreateOutputFile(outputFile string) (*os.File, error) {
+	if _, err := os.Stat(outputFile); os.IsNotExist(err) {
+		return os.Create(outputFile)
+	}
+	return nil, fmt.Errorf("output file %s already exists", outputFile)
```
peanball, line 239: Why is this an error? Do we need some logic to force overwrite where this does not kick in?

src/pcap/pcap_test.go
```diff
@@ -4,17 +4,17 @@ import (
...
 	"net"
 	"strconv"
 	"sync"
 	"testing"
-
-	"github.com/go-playground/validator/v10"
-	"go.uber.org/zap"
-	"go.uber.org/zap/zaptest/observer"
-	"google.golang.org/grpc/metadata"
 )
```
peanball, line 18: The linter will have a few words to say about this.

src/pcap/bosh/bosh.go
```diff
@@ -0,0 +1,209 @@
...
+	AccessToken     string       `yaml:"access_token"`
+	AccessTokenType string       `yaml:"access_token_type"`
+	Alias           string       `yaml:"alias"`
+	CaCert          string       `yaml:"ca_cert"`
+	RefreshToken    string       `yaml:"refresh_token"`
+	RawDirectorURL  string       `yaml:"url"`
+	DirectorURL     *url.URL     `yaml:"-"`
+	UaaURL          *url.URL     `yaml:"-"`
+	client          *http.Client `yaml:"-"`
+}
```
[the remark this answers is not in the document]
a18e: yes otherwise they will be written to file

src/pcap/cmd/mock-bosh-api/main.go
a18e: will keep for now, will be removed before release along with agent-cli and api-cli

src/pcap/bosh/bosh.go
```diff
@@ -0,0 +1,209 @@
...
+			Status bool `json:"status"`
+		} `json:"snapshots"`
+		ConfigServer struct {
+			Status bool `json:"status"`
+			Extras struct {
+				Urls []string `json:"urls"`
+			} `json:"extras"`
+		} `json:"config_server"`
+	} `json:"features"`
+}
```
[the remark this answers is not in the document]
a18e: Reduced number of fields in the new `pcap.BoshInfo`

src/pcap/client.go
```diff
@@ -0,0 +1,240 @@
...
+}
+
+func (c *Client) HandleRequest(endpointRequest *EndpointRequest, options *CaptureOptions, ctx context.Context, cancel CancelCauseFunc) error {
+	// setup output/pcap-file
+	packetWriter := pcapgo.NewWriter(c.packetFile)
+	err := packetWriter.WriteFileHeader(options.SnapLen, layers.LinkTypeEthernet)
+	if err != nil {
+		return err
+	}
+
```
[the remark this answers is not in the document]
a18e: I added a force overwrite option and added a check if the filepath is valid.
Why did we want to write the header only after the first packet arrives? We would need to do some refactoring...

src/pcap/cmd/common.go
```diff
@@ -1,125 +0,0 @@
...
-	"google.golang.org/grpc"
-	"google.golang.org/grpc/codes"
-	"google.golang.org/grpc/credentials"
-	"google.golang.org/grpc/credentials/insecure"
-	"google.golang.org/grpc/status"
-)
-
-var zapConfig zap.Config
-
-func init() {
```
b1tamara, line 23: If you remove the file incl. init-function where the log was configured, you need to configure that again in pcap-agent/main.go and pcap-api/main.go. 

Refer to the commit https://github.com/sap-contributions/pcap-release/commit/dbbdb4107857347a5d699655448cb14e641de28f#diff-6eb5e658f7d5bfc4b5feb25f7207fcac7cd96c459ccf2a6210ca13661481f705R20

src/pcap/bosh.go
```diff
@@ -1,55 +1,414 @@
...
+					ClientAuth: tls.RequireAndVerifyClientCert,
+					RootCAs:    br.boshRootCAs,
+				},
+			},
+		}
+	}
+
+	br.logger.Info("discovering bosh-UAA endpoint")
+	ptr := *br.directorURL
+	infoEndpoint := &ptr
```
b1tamara, line 184: Do we need this ptr variable? It is used only for assignment to infoEndpoint.

src/pcap/bosh.go
```diff
@@ -1,55 +1,414 @@
...
+
+	if len(boshRequest.Groups) == 0 {
 		return fmt.Errorf("invalid message: instance group(s): %w", errEmptyField)
 	}
 
 	return nil
 }
+
+func (br *BoshResolver) setup() error {
+	br.logger.Info("setting Up BoshResolver", zap.Any("resolver-config", br.config))
```
b1tamara, line 165: "setting up BoshResolver": Up -> up

src/pcap/bosh.go
```diff
@@ -1,55 +1,414 @@
...
-func (bosh *BoshHandler) name() string {
-	return "bosh"
+func (br *BoshResolver) Name() string {
+	return fmt.Sprintf("bosh/%s", br.config.EnvironmentAlias)
 }
 
-func (bosh *BoshHandler) canResolve(request *Capture) bool {
-	return request.GetBosh() != nil
+func (br *BoshResolver) CanResolve(request *EndpointRequest) bool {
+	if boshRequest := request.GetBosh(); boshRequest != nil {
```
b1tamara, line 98: Is it possible that incoming _request_ is nil?

src/pcap/bosh.go
```diff
@@ -1,55 +1,414 @@
...
+//   - for RSA signed JWT that the RSA signature is consistent with the key provided by UAA
+//   - that there is a claim 'scope' that contains one entry that matches neededScope.
+//
+// Limitations: only RSA signed tokens are supported.
+//
+// returns a boolean that confirms that the token is valid, from a valid issuer and has the needed scope,
+// and an error in case anything went wrong while verifying the token and its scopes.
+func (br *BoshResolver) verifyJWT(tokenString string) (bool, error) {
+	token, err := jwt.Parse(tokenString, func(token *jwt.Token) (interface{}, error) {
+		if jku, ok := token.Header["jku"]; ok {
```
b1tamara, line 295: To get the function more readable, it is the better way to check if ok is false and return an error. Like
`jku, ok := token.Header["jku"]
  if !ok {
	return nil, fmt.Errorf("header 'jku' missing from token, cannot verify signature")
  }
  jkuURL, err := url.Parse(jku.(string))`

src/pcap/bosh.go
```diff
@@ -1,55 +1,414 @@
...
+// (key ID, RSA public key), which is used to verify the token.
+//
+// Limitation: only supports RSA tokens using the 'jku' header, which points to a URL
+// that can be used to retrieve key information.
+func (br *BoshResolver) parseRsaToken(token *jwt.Token) (interface{}, error) {
+	if rsa, ok := token.Method.(*jwt.SigningMethodRSA); ok {
+		// with the RSA signing method, the key is a public key / certificate that can be
+		// retrieved from the JKU endpoint (among other places).
+		if rawKeyInfoURL, ok := token.Header["jku"].(string); ok {
+			var kid string
```
b1tamara, line 346: The same as above. I would prefer to check if it is not ok and return in this case. And continue with non-error flow.

src/pcap/bosh.go
```diff
@@ -1,55 +1,414 @@
...
+	}
+
+	err = json.Unmarshal(data, &keys)
+	if err != nil {
+		return nil, err
+	}
+
+	for _, key := range keys.Keys {
+		if key.Kty == "RSA" && key.Kid == kid {
+			matchingKey := key
```
b1tamara, line 408: Do we really need a new variable?

src/pcap/bosh.go
```diff
@@ -1,55 +1,414 @@
...
+
+		return nil, fmt.Errorf("could not find key information URL in token headers: %+v", token.Header)
+	}
+
+	return nil, fmt.Errorf("unsupported signing method: %v", token.Header["alg"])
+}
+
+// fetchPublicKey fetches the token key information from url and returns the key with the Key ID (kid).
+//
+// returns an error if no key can be found with the requested kid or an error arises while communicating with url.
```
b1tamara, line 377: Capitalize "returns" as it is the first word of the sentence.

src/pcap/client.go
```diff
@@ -0,0 +1,240 @@
...
+}
+
+func (c *Client) ConnectToAPI(apiURL *url.URL) error {
+	var (
+		err   error
+		creds credentials.TransportCredentials
+	)
+
+	if apiURL.Scheme == "https" {
+		creds, err = LoadTLSCredentials("", "", nil, nil, nil)
```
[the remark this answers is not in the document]
b1tamara: at least for TLS you can transfer here peer CAs and serverName.

src/pcap/api.go
```diff
@@ -98,9 +94,19 @@ func (api *API) Status(context.Context, *StatusRequest) (*StatusResponse, error)
...
+		resolverNames[i] = key
+		i++
+	}
+	return resolverNames
+}
+
+// resolverRegistered checks if handler is registered.
 // returns false, if the handler is not registered.
-func (api *API) handlerRegistered(handler string) *bool {
+func (api *API) resolverRegistered(handler string) *bool {
```
b1tamara, line 109: I am not sure the terms handler and resolver in the api.go are used in the consistent way.

src/pcap/api.go
```diff
@@ -98,9 +94,19 @@ func (api *API) Status(context.Context, *StatusRequest) (*StatusResponse, error)
...
+	i := 0
+	for key := range api.resolvers {
+		resolverNames[i] = key
+		i++
+	}
+	return resolverNames
+}
+
+// resolverRegistered checks if handler is registered.
 // returns false, if the handler is not registered.
```
b1tamara, line 108: Capitalize "returns" as it is the first word in the sentence.

src/pcap/client.go
```diff
@@ -0,0 +1,259 @@
...
+	case MessageType_LIMIT_REACHED:
+		logLevel = zapcore.WarnLevel
+	case MessageType_CAPTURE_STOPPED:
+		logLevel = zapcore.InfoLevel
+	case MessageType_CONNECTION_ERROR:
+		logLevel = zapcore.ErrorLevel
+	}
+	// TODO: Find a way to assign the proper log level.
+	_ = logLevel
+	zap.S().Infof("%s(%s): %s", message.Origin, message.Type, message.Message)
```
b1tamara, line 213: We agreed do not use the sugared logger.

src/pcap/cmd/common.go
peanball: The logging initialisation is now in `pcap/logger.go`, not removed. It's still loaded automatically for any process that pulls in the `pcap` module.

src/pcap/bosh.go
peanball: I will add a check. But it will just return `false` in that case because no resolver can resolve an empty request.

src/pcap/bosh.go
peanball: Yes, fair.

src/pcap/bosh.go
```diff
@@ -1,55 +1,414 @@
...
+					return br.parseRsaToken(token)
+				}
+			}
+			return nil, fmt.Errorf("header 'jku' %v did not match any UAA base URLs reported by the BOSH Director: %v", jku, br.uaaURLs)
+		}
+		return nil, fmt.Errorf("header 'jku' missing from token, cannot verify signature")
+	})
+
+	if err != nil || !token.Valid {
+		return false, err
```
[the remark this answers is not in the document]
peanball: yes. Slipped through the PoC review back then.

src/pcap/bosh.go
peanball: My idea was to have a copy of the one value I need, so the full response can be garbage collected.

src/pcap/client.go
peanball: This is for use in the CLI, different rules apply.
But realistically, the `Client` should not log anything itself. I proposed some 'callback handler' that gets packets or messages and can decide what to do with those. The callback handler would be defined in the app that uses the client. Then that app can define how to log / format outputs.

src/pcap/client.go
```diff
@@ -0,0 +1,259 @@
...
+	}
+
+	err = c.stream.Send(captureRequest)
+	if err != nil {
+		return err
+	}
+
+	copyWg := &sync.WaitGroup{}
+	copyWg.Add(1)
+	go handleStream(c.stream, packetWriter, copyWg, cancel)
```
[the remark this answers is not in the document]
peanball: Added now. Also identified when doing the bosh resolver integration tests / scenarios.

src/pcap/bosh.go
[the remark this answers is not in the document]
peanball: We can then leave out the boolean altogether. If there is no error, the token is valid. If something is wrong, the error tells you what it is.

src/pcap/client.go
```diff
@@ -0,0 +1,335 @@
...
+	return client, nil
+}
+
+func (c *Client) Stop() {
+	c.StopRequest()
+}
+
+func (c *Client) Wait() {
+	// waiting is implicit. Needed for fulfulling the pcap.Stoppable interface.
+}
```
maxmoehl, line 106: This seems very wrong, it makes a promise that it cannot fulfill. If the interface is too strict we should loosen it or make the client support this functionality.

maxmoehl (changes requested):
Two things that are not explicitly mentioned:
- What is the docker stuff for and how can it be used?
- There could be more documentation in the form of godoc comments

src/pcap/cmd/pcap-bosh-cli/main.go
```diff
@@ -0,0 +1,472 @@
...
+		},
+		Header: http.Header{
+			"Accept": {"application/json"},
+		},
+	})
+	if err != nil {
+		return fmt.Errorf("could not get response from bosh-director %w", err)
+	}
+
+	defer pcap.CloseQuietly(res.Body)
```
maxmoehl, line 410: Please consider using `defer func() { _ = res.Body.Close() }()`

src/pcap/bosh.go
```diff
@@ -1,55 +1,512 @@
...
+	br.UaaURLs = apiResponse.UserAuthentication.Options.URLs
+	br.logger.Info("connected to bosh-director", zap.Any("bosh-director", br.DirectorURL.String()))
+	return nil
+}
+
+// info retrieves the BOSH director /info endpoint.
+//
+// Used for startup and health check.
+func (br *BoshResolver) info() (*BoshInfo, error) {
+	// FIXME: Workaround for URL.JoinPath, which is buggy: https://github.com/golang/go/issues/58605
```
maxmoehl, line 252: Does it work if we set the initial path after parsing to `/`?

src/pcap/cmd/pcap-bosh-cli/main.go
```diff
@@ -0,0 +1,472 @@
...
+}
+
+// fetchUAAURL connects to the bosh-director API to fetch the bosh-uaa API URL.
+func (e *Environment) fetchUAAURL() error {
+	res, err := e.client.Do(&http.Request{ //nolint:bodyclose // closed via pcap.CloseQuietly below.
+		Method: http.MethodGet,
+		URL: &url.URL{
+			Scheme: e.DirectorURL.Scheme,
+			Host:   e.DirectorURL.Host,
+			Path:   "/info",
```
maxmoehl, line 400: Consider using `JoinPath` if that works (see other comment)

src/pcap/devtools/reader.go
```diff
@@ -0,0 +1,40 @@
...
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+)
+
+type genericStreamReceiver interface {
+	Recv() (*pcap.CaptureResponse, error)
+}
+
+// ReadN reads a number of messages from stream
+// TODO: Remove this when we have a proper CLI
```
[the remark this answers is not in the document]
peanball: It can stay, but realistically there's no strong need for the devtools. We can decide to remove them a bit later.

src/pcap/cmd/pcap-api/main.go
```diff
@@ -33,42 +32,62 @@ func main() {
 	}
 
 	if err != nil {
-		log.Fatal("unable to initialize", zap.Error(err))
+		log.Panic("Unable to initialize", zap.Error(err))
```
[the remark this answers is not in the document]
peanball: The only remaining panics are now in test code where it's ok and in the dev tools that should eventually be removed as we now have proper clients for things.

I will leave dev tools around though as they _may_ still be useful while developing the next bits.

src/pcap/cmd/pcap-api/main.go
```diff
@@ -33,42 +32,62 @@ func main() {
...
 }
+
+// registerBoshResolvers tries to register all BoshResolvers defined in configs and registers them in api.
+//
+// Logs an error if the resolver could not be initialized.
+func registerBoshResolvers(configs []pcap.BoshResolverConfig, log *zap.Logger, api *pcap.API) {
+	for _, env := range configs {
+		resolver, err := pcap.NewBoshResolver(env)
+		if err != nil {
+			log.Panic("Failed to setup BoshResolver", zap.String(pcap.LogKeyResolver, env.EnvironmentAlias), zap.Error(err))
```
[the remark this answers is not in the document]
peanball: The remaining question is: is our assumption that anything in the config _must_ connect and the API will not start otherwise? If so, I'd happily bubble up the error and say there that init failed because we couldn't connect to BOSH env xzy.

src/pcap/cmd/pcap-bosh-cli/main.go
peanball: Done.

src/pcap/bosh.go
peanball: It does.

src/pcap/client.go
```diff
@@ -0,0 +1,335 @@
...
+	if err != nil {
+		c.log.Error("could not stop")
+	}
+	c.stopped = true
+}
+
+// ReadCaptureResponse reads CaptureResponse's from the api in a loop and delegates writing/logging messages & packets to WriteMessage / writePacket.
+//
+// It terminates if an error or clean stop-message is received.
+func (c *Client) ReadCaptureResponse(stream API_CaptureClient, packetWriter *pcapgo.Writer, copyWg *sync.WaitGroup, cancel context.CancelCauseFunc) {
```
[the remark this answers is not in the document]
peanball: @maxmoehl could I not also pass in the context and close (cancel w/o error) that context?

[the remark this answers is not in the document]
maxmoehl: It's the other way around. The caller wants to be sure that this function stopped after the context was already cancelled. This cannot be achieved with a context because the context does not guarantee that the work is actually cancelled which is why we need a separate control to ensure the work has stopped.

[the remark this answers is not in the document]
maxmoehl: The go idiomatic way would (IMO) be a channel since it's only one routine in the wait group. Technically both approaches work just fine.

src/pcap/cmd/pcap-api/main.go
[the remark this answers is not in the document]
maxmoehl: Right now this is the assumption that we make. If we decide that this is not the case you would have to add logic to check if at least one resolver works which adds more complexity. Let's keep it simple for now and fail if something in the config is unavailable.

src/pcap/client.go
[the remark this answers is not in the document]
peanball: channel it is then! thank you!

src/pcap/client.go
```diff
@@ -0,0 +1,335 @@
...
+	err := packetWriter.WritePacket(captureInfo, packet.Data)
+	if err != nil {
+		log.Error("writing packet to file failed", zap.Error(err))
+	}
+	if log.Level().Enabled(zap.DebugLevel) {
+		log.Debug("received packet", zap.Int("bytes", len(packet.Data)), zap.Time("capture-timestamp", packet.Timestamp.AsTime()))
+	}
+}
+
+// logProgress logs out the size of the outputFile every 5 seconds (see logProgressWait).
```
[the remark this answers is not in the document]
peanball: There is no good reason for this. Logs are all directed to stderr anyway.

src/pcap/client.go
```diff
@@ -0,0 +1,335 @@
...
+				return
+			}
+			logger.Debug(fmt.Sprintf("%s bytes written to disk (total).", bytefmt.ByteSize(uint64(info.Size()))))
+		case <-ctx.Done():
+			return
+		}
+	}
+}
+
+// CheckAPIHandler checks if handler is announced by the API as healthy available handler.
```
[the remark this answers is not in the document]
peanball: overzealous refactor in IntelliJ

src/pcap/client.go
peanball: I will instead remove the Stoppable interface from the client, as it doesn't fit there.

src/pcap/api.go
```diff
@@ -183,9 +199,11 @@ func (api *API) Capture(stream API_CaptureServer) (err error) {
 		return errorf(codes.InvalidArgument, "expected start message, got %v: %w", req.Operation, errUnexpectedMessage)
 	}
 
-	targets, resolveErr := api.resolveAgentEndpoints(opts.Start.Capture, log)
-	if errors.Is(resolveErr, errValidationFailed) {
-		return errorf(codes.InvalidArgument, "capture targets not found: %w", err)
+	targets, resolveErr := api.resolveAgentEndpoints(opts.Start.Request, log)
+	if errors.Is(resolveErr, ErrValidationFailed) {
+		return errorf(codes.InvalidArgument, "capture targets not found: %w", resolveErr)
```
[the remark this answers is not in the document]
maxmoehl: I'm struggling to understand what I wrote there :joy:

But looking at it again the confusing bit seems to be that `InvalidArgument` and `NotFound` are two separate things even though they are related. I'm saying the message to the client should be "I can't find what you are looking for" instead of "Your request is malformed" which is expressed in the text but not in the status code.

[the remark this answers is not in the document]
maxmoehl: This remark might be a bit late but I will still raise it: Maybe the base error codes should already have the status code assigned to them? After all the bottom most error code usually gives the underlying reason and has the best idea of what exactly went wrong. The mapping seems pretty obvious:
```
ErrValidationFailed  = fmt.Errorf("validation failed")            // InvalidArgument
errTooManyCaptures   = fmt.Errorf("too many concurrent captures") // ResourceExhausted
errDraining          = fmt.Errorf("draining")                     // Unavailable
errUnexpectedMessage = fmt.Errorf("unexpected message")           // InvalidArgument
ErrNoEndpoints       = fmt.Errorf("no matching endpoints found")  // NotFound
ErrResolverUnhealthy = fmt.Errorf("resolver unhealthy")           // Internal
ErrNotAuthorized     = fmt.Errorf("not authorized")               // PermissionDenied (or maybe Unauthenticated, not sure)
```

src/pcap/cmd/pcap-bosh-cli/main.go
```diff
@@ -0,0 +1,472 @@
...
+	err := e.refreshTokens()
+	if err != nil {
+		return fmt.Errorf("failed to refresh bosh access token %w", err)
+	}
+	return nil
+}
+
+// init sets up a Bosh environment by parsing the bosh-director URL (string) from the config-file
+// and then sets up the http client for use with either TLS or plain HTTP.
+func (e *Environment) init() error {
```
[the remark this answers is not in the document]
maxmoehl: You could still do that by passing a reader (or string) into `NewEnvironment` and delegating the parsing to it, the initialisation can be done after parsing in the same function.

src/pcap/api.go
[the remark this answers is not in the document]
peanball: We will tackle this in a different PR in the future.

src/pcap/config.go
```diff
@@ -50,7 +61,88 @@ type Listen struct {
...
+	ID       string     `yaml:"id" validate:"required"`
+}
+
+// TLSCredentials creates the necessary credentials from this Config. If NodeConfig.Listen.TLS is
+// nil, credentials which disable transport security, will be used.
+//
+// Note: the TLS version is currently hard-coded to TLSv1.3.
+func (c NodeConfig) TLSCredentials() (credentials.TransportCredentials, error) {
+	if tlsConfig := c.Listen.TLS; tlsConfig != nil {
+		return LoadTLSCredentials(tlsConfig.Certificate, tlsConfig.PrivateKey, &tlsConfig.CertificateAuthority, nil, nil)
```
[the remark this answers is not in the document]
peanball: Not via the NodeConfig. We're starting the next stream to finalise the 'boshrelease' part with a full deployment including mTLS between all necessary components. Expanding should be best done there, as that task will at the same time be able to fully verify the changes.

We have functionality to load all the necessary data, and the configuration data structures to express it.

src/pcap/client.go
```diff
@@ -0,0 +1,335 @@
...
+	log           *zap.Logger
+	stream        API_CaptureClient
+	messageWriter MessageWriter
+	stopped       bool
+	aPIClient
+}
+
+func CloseQuietly(closer io.Closer) {
+	_ = closer.Close()
+}
```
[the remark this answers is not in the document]
peanball: decided to remove closeQuietly altogether.

src/pcap/api.go
```diff
@@ -98,11 +101,24 @@ func (api *API) Status(context.Context, *StatusRequest) (*StatusResponse, error)
 	return apiStatus, nil
 }
 
-// handlerRegistered checks if handler is registered.
+func (api *API) RegisteredResolverNames(onlyHealthy bool) []string {
```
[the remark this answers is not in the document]
peanball: It's gone. The function now only lists healthy resolvers. We don't really care about the others anyway and crash if we don't have any at startup (to be restarted by the surrounding task management system)

src/pcap/cmd/pcap-bosh-cli/main.go
```diff
@@ -0,0 +1,455 @@
...
+	if err != nil {
+		return err
+	}
+	e.UaaURL = uaaURL
+
+	return nil
+}
+
+// refreshTokens connects to the bosh-uaa API to fetch updated bosh access- & refresh-token.
+func (e *Environment) refreshTokens() error { //TODO: logging
```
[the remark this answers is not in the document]
peanball: It's all handled via errors. No need for extra logging.

src/pcap/bosh.go
```diff
@@ -1,55 +1,512 @@
...
+
+	err = json.Unmarshal(data, &keys)
+	if err != nil {
+		return nil, err
+	}
+
+	for _, key := range keys.Keys {
+		if key.Kty == "RSA" && key.Kid == kid {
+			matchingKey := key
+			return &matchingKey, nil
```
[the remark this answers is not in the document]
peanball: Maybe that was premature optimization. The idea was to copy the key and not have to retain `keys`, so it could be collected. Possibly not really needed. @b1tamara already mentioned this as well.

src/pcap/cmd/pcap-bosh-cli/main.go
```diff
@@ -0,0 +1,472 @@
...
+		DisableStacktrace: true,
+		Encoding:          "console",
+		EncoderConfig:     encoderConfig,
+		OutputPaths:       outputPaths,
+		ErrorOutputPaths:  outputPaths,
+	}
+	logger = zap.Must(zapConfig.Build())
+	zap.ReplaceGlobals(logger)
+	logger.Debug("successfully set up logger", zap.Strings("output-paths", outputPaths))
+}
```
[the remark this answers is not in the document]
peanball: With minor differences. As discussed I will add it to our list of future improvements to have something that starts initializing the logger _after_ we've read the config and know what we actually want (e.g. log level, outputs paths, etc.)

src/pcap/cmd/pcap-bosh-cli/main.go
```diff
@@ -0,0 +1,472 @@
...
+	err := c.CheckAPIHandler(fmt.Sprintf("bosh/%v", environmentAlias))
+	return fmt.Errorf("pcap-api does not support BOSH environment %v, %w", environmentAlias, err)
+}
+
+// setLogLevel sets the log level of the zap.Logger created in setupLogging via atomicLogLevel
+//
+// It accepts bool values of the verbose and quiet options.
+//
+// verbose sets the loglevel to Debug, quiet to Warn. The default is Info. If both verbose and quiet are specified, we return an error.
+func setLogLevel(verbose bool, quiet bool) error {
```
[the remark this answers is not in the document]
peanball: same as above, setting the log level after the fact is not great for an initialization. To be done as future improvement.

src/pcap/bosh.go
```diff
@@ -1,55 +1,512 @@
...
+			return key, err
+		}
+
+		return nil, fmt.Errorf("could not find key information URL in token headers: %+v", token.Header)
+	}
+
+	return nil, fmt.Errorf("unsupported signing method: %v", token.Header["alg"])
+}
+
+func (br *BoshResolver) verifyRSASignature(token *jwt.Token, rsa *jwt.SigningMethodRSA) (interface{}, bool, error) {
```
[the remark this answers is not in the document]
peanball: I've added `*rsa.PublicKey` as far in the hierarchy as I could. The final call  to the callback function from the JWT library must have `interface{}` in the signature.

peanball:
@maxmoehl 
> What is the docker stuff for and how can it be used?

Explained now in [8b2db46](https://github.com/cloudfoundry/pcap-release/pull/17/commits/8b2db46a502e7196097ff6fbd3ecff5897c1fb88).

> There could be more documentation in the form of godoc comments
This should also be taken care of now.

src/pcap/cmd/pcap-bosh-cli/main.go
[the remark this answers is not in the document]
peanball: I chose the way of giving an initialized environment when querying for it. This is localized into the pcap-bosh-cli and not general pcap code.

src/pcap/bosh.go
```diff
@@ -1,55 +1,512 @@
...
+		jkuURL, err := url.Parse(jku.(string))
+		if err != nil {
+			return nil, err
+		}
+
+		for _, issuer := range br.UaaURLs {
+			var issuerURL *url.URL
+			issuerURL, err = url.Parse(issuer)
+			if err != nil {
+				br.logger.Warn("could not parse URL %s: %v", zap.String("issuer", issuer), zap.Error(err))
```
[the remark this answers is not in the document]
peanball: The "URL" is `issuer`. So it's there.

src/pcap/bosh.go
```diff
@@ -1,55 +1,512 @@
...
+	err = json.Unmarshal(data, &apiResponse)
+	if err != nil {
+		return nil, fmt.Errorf("could not parse bosh-director API response: %w", err)
+	}
+	return apiResponse, nil
+}
+
+// Healthy returns true if the resolver ran setup() and can connect to the BOSH director.
+func (br *BoshResolver) Healthy() bool {
+	if br.client == nil {
```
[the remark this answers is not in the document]
peanball: As discussed, panicking in case setup was not called (which requires you to instantiate your own `BoshResolver` and not call `NewBoshResolver`) is OK behaviour. So I will remove these checks.

src/pcap/bosh.go
```diff
@@ -1,55 +1,512 @@
...
+
+// info retrieves the BOSH director /info endpoint.
+//
+// Used for startup and health check.
+func (br *BoshResolver) info() (*BoshInfo, error) {
+	// FIXME: Workaround for URL.JoinPath, which is buggy: https://github.com/golang/go/issues/58605
+	infoEndpoint := *br.DirectorURL
+	infoEndpoint.Path = "/info"
+
+	response, err := br.client.Do(&http.Request{
```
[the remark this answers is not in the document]
peanball: As discussed, setup() is called in NewBoshResolver. Failing to call it will result in a panic, but is also wrong use of the API.

src/pcap/process.go
```diff
@@ -0,0 +1,55 @@
...
+	Stoppable
+
+	Wait()
+}
+
+// WaitForSignal to tell the process to stop processing any streams. Will first tell the agent
+// to end any running streams, wait for them to terminate and gracefully stop the gRPC server
+// afterwards. Currently listens for SIGUSR1 and SIGINT, SIGQUIT and SIGTERM.
+//
+// the provided Stoppable can also be a WaitingStoppable. Then the Wait() function is also called.
```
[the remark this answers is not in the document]
peanball: Rewrote the function to support specific signals defined at the call site and adjusted the godocs that this is a generic function to stop things via signals.

src/pcap/cmd/pcap-bosh-cli/main_test.go
```diff
@@ -0,0 +1,56 @@
+package main
```
maxmoehl, line 1: Could use a few more tests, but this can also be tackled in a follow-up PR.

src/pcap/test/integration/agent_api_client.go
```diff
@@ -0,0 +1,445 @@
...
+	. "github.com/onsi/ginkgo/v2" //nolint:revive,stylecheck // this is the common way to import ginkgo and gomega
+	. "github.com/onsi/gomega"    //nolint:revive,stylecheck // this is the common way to import ginkgo and gomega
+	"go.uber.org/zap"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/codes"
+)
+
+var apiClient pcap.APIClient
+
+var MaxConcurrentCaptures = uint(2)
```
maxmoehl, line 24:
```suggestion
var MaxConcurrentCaptures uint = 2
```

src/pcap/api.go
```diff
@@ -224,31 +243,35 @@ func (api *API) Capture(stream API_CaptureServer) (err error) {
...
-func loadTLSCredentials(agents AgentMTLS) (credentials.TransportCredentials, error) {
-	if agents.MTLS == nil || agents.MTLS.SkipVerify {
+func loadTLSCredentials(agentMTLS *MutualTLS) (credentials.TransportCredentials, error) {
+	if agentMTLS == nil || agentMTLS.SkipVerify {
 		return insecure.NewCredentials(), nil
 	}
-	mTLS := agents.MTLS
+
+	mTLS := *agentMTLS
 	return LoadTLSCredentials(mTLS.Certificate, mTLS.PrivateKey, nil, &mTLS.CertificateAuthority, &mTLS.CommonName)
```
maxmoehl, line 252:
```suggestion
	return LoadTLSCredentials(agentMTLS.Certificate, agentMTLS.PrivateKey, nil, &agentMTLS.CertificateAuthority, &agentMTLS.CommonName)
```

src/pcap/bosh.go
```diff
@@ -1,55 +1,553 @@
...
+
+	var apiResponse *BoshInfo
+	data, err := io.ReadAll(response.Body)
+	if err != nil {
+		return nil, fmt.Errorf("could not read bosh-director API response: %w", err)
+	}
+	err = json.Unmarshal(data, &apiResponse)
+	if err != nil {
+		return nil, fmt.Errorf("could not parse bosh-director API response: %w", err)
+	}
```
maxmoehl, line 302: Please use `json.NewDecoder(response.Body).Decode(&apiResponse)`.

But this shouldn't work anyway because the apiResponse pointer is nil since it has never been initialized and you are taking a pointer to a pointer.

src/pcap/bosh.go
```diff
@@ -1,55 +1,553 @@
...
+		return nil, err
+	}
+
+	defer func() { _ = res.Body.Close() }()
+
+	keys := struct {
+		Keys []UaaKeyInfo
+	}{}
+
+	data, err := io.ReadAll(res.Body)
```
maxmoehl, line 536: use `json.NewDecoder`, see above

maxmoehl (changes requested):
Just a few things left :)

src/pcap/cmd/pcap-bosh-cli/main.go
```diff
@@ -0,0 +1,488 @@
...
+	if err != nil {
+		return
+	}
+
+	logger.Debug("bosh-config and tokens successfully updated")
+
+	// set up pcap-client/pcap-api connection
+	client, err = pcap.NewClient(opts.File, logger, pcap.LogMessageWriter{Log: logger})
+	if err != nil {
+		err = fmt.Errorf("could not set up pcap-client %w", err)
```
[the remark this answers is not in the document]
peanball: Went through all of them and ensured that it's correct now. Also found some gems like trying to wrap `zap.Error(err)`.

src/pcap/cmd/pcap-bosh-cli/main_test.go
peanball: Yes. We still want to build acceptance tests in a follow-up story. The CLI was planned to be tested in depth there at the latest.

src/pcap/cmd/pcap-bosh-cli/main.go
```diff
@@ -0,0 +1,488 @@
...
+		environment *Environment
+	)
+
+	// Parse command-line arguments
+	_, err = flags.Parse(&opts)
+	if err != nil {
+		os.Exit(1)
+	}
+
+	defer func() {
```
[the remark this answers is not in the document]
peanball: This was on purpose. I didn't want to have double outputs of the flags.Parse error, because flags.Parse already prints the error message itself.

I couldn't figure out in reasonable time how to distinguish the flags error cleanly enough to not print it conditionally.

I would add a comment on the first exit to explain why it's there.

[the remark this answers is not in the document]
peanball: Nevermind, figured it out. Will be in the next commit.

src/pcap/cmd/pcap-api/main.go
```diff
@@ -33,42 +40,72 @@ func main() {
 	}
 
 	if err != nil {
-		log.Fatal("unable to initialize", zap.Error(err))
+		log.Error("Unable to initialize", zap.Error(err))
```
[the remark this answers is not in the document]
peanball: added

src/pcap/cmd/pcap-bosh-cli/main.go
```diff
@@ -0,0 +1,488 @@
...
+}
+
+// checkAPIHealth accepts a Client with working client-connection and an environmentAlias.
+//
+// Using the clients connection to the pcap-api it checks whether the api endpoint is healthy in general
+// and if it supports requests to the Bosh Environment specified in environmentAlias.
+func checkAPIHealth(c *pcap.Client, environmentAlias string) error {
+	err := c.CheckAPIHandler(fmt.Sprintf("bosh/%v", environmentAlias))
+	return fmt.Errorf("pcap-api does not support BOSH environment %v, %w", environmentAlias, err)
+}
```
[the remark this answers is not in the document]
peanball: Good catch, and another reason why I should have pushed for CLI tests more.

maxmoehl (approved):

PR #19, opened 2023-04-14
Remove support for multiple BOSH Environments

peanball:
There is no good use case to support multiple environments and it makes implementing the boshrelease configuration unnecessarily complex.


src/pcap/.golangci.yml
```diff
@@ -286,3 +285,6 @@ issues:
     - gosec
     - noctx
     - wrapcheck
+#  - path: "\\.pb\\.go"
+#    linters: []
+
```
maxmoehl, line 290: Seems like something sneaked in here that shouldn't.

maxmoehl (changes requested):

src/pcap/.golangci.yml
peanball: yes it did. fixed now.

maxmoehl (approved):

src/pcap/cmd/pcap-bosh-cli/main.go
```diff
@@ -298,18 +298,17 @@ func connectToEnvironment(environmentAlias string, config *Config) (*Environment
 			return &environment, nil
 		}
 	}
-	return nil, fmt.Errorf("could not find bosh-environment %s in config", environmentAlias)
+	return nil, fmt.Errorf("could not find bosh-environment %s in BOSH CLI config", environmentAlias)
```
b1tamara, line 301: we have to check if the whole function is still needed

peanball: Yes because you still need to connect to the one environment.

PR #22, opened 2023-04-21
build: Update blobs

b1tamara:
Add blob for golang 1.20 instead of 1.18


config/blobs.yml
```diff
@@ -2,6 +2,10 @@ golang/go1.18.5.linux-amd64.tar.gz:
   size: 141855946
   object_id: d380aecf-44d1-4003-61f5-36a76ad4c864
   sha: sha256:9e5de37f9c49942c601b191ac5fba404b868bfc21d446d6960acc12283d6e5f2
+golang/go1.20.2.linux-amd64.tar.gz:
```
peanball, line 5: 1.20.3 is available since a while. Maybe we should use that instead.

peanball (changes requested):
update to 1.20.3 please.

peanball (approved):

PR #23, opened 2023-04-27
Fixes for boshrelease

a18e:
boshrelease:
- allow pcap-api to run w/o TLS to platform (TLS currently not working)
-- add spec tests for non-TLS mode 
-- adjust example manifest/ops files to run w/o platform TLS
- add pcap-agent to pcap-api manifest so it get's deployed on the pcap-api VMs as well
- update dependencies: stemmcell, releases
- mTLS between api & agents currently works using a workaround: pcap-agents reuse whole cert-chain (instead of just CA) from pcap-api
- minimal supported TLS version is now 1.2

pcap-api Go code:
- fix bug with decoding certificate files with newlines
- add unit tests for certificate file parsing


jobs/pcap-api/templates/pcap-api.yml.erb
```diff
@@ -10,15 +10,20 @@ config = {
...
-    "tls" => {
-      "certificate"=> "/var/vcap/jobs/pcap-api/config/certs/pcap-api.crt",
-      "private_key" => "/var/vcap/jobs/pcap-api/config/certs/pcap-api.key",
-      "ca" => "/var/vcap/jobs/pcap-api/config/certs/pcap-api-ca.crt",
-    },
   },
   "cli_download_root" => p("pcap-api.cli_download_root")
 }
 
+if_p("pcap-api.listen.tls.certificate", "pcap-api.listen.tls.private_key", "pcap-api.listen.tls.ca") do
```
maxmoehl, line 17: As much as I hate the `enabled` property for such things, after careful consideration I suggest we do use it here for a few reasons:
- This will fail in unexpected ways, e.g. you don't provide a CA and get no TLS at all without any error.
- We cannot check if the parent node `pcap-api.listen.tls` exists (thanks to bosh).
- With the `enabled` property we will explicitly fail if it is `true` but one of the required fields is missing.

WDYT?

(sorry that I suggested the opposite yesterday)

maxmoehl (changes requested):

jobs/pcap-api/templates/pcap-api.yml.erb
a18e: I agree, but want to get more feedback before implementing.

manifests/pcap-api.yml
```diff
@@ -39,15 +39,34 @@ instance_groups:
...
+              size: 1000
+              upper_limit: 995
+              lower_limit: 900
+            concurrent_captures: 5
+            listen:
+              port: 8080
+            #              tls:
+            #                certificate: ((pcap_api_platform_mtls.certificate))
+            #                private_key: ((pcap_api_platform_mtls.private_key))
+            #                ca: ((/bosh-aws-cfn07/cf/router_ca.ca))
```
b1tamara, line 53: Adapt this lines

manifests/pcap-api.yml
```diff
@@ -39,15 +39,34 @@ instance_groups:
...
+            bosh:
+              director_url: ((pcap_api.bosh_director_api))
+              token_scope: bosh.admin
+              agent_port: 9494
+              mtls:
+                common_name: bosh.service.cf.internal
+                skip_verify: false
+                certificate: ((pcap_api_bosh_mtls.certificate))
+                private_key: ((pcap_api_bosh_mtls.private_key))
+                ca: ((/bootstrap-bosh-aws-cfn07/bosh/default_ca.ca))
```
b1tamara, line 63: Please adapt the path for the ca property

manifests/pcap-api.yml
```diff
@@ -78,36 +114,67 @@ update:
...
+  - name: pcap_api_bosh_mtls
+    type: certificate
+    options:
+      ca: /bootstrap-bosh-aws-cfn07/bosh/default_ca
+      common_name: pcap_api_bosh_mtls
+      alternative_names:
+        - pcap-api.service.cf.internal
+      extended_key_usage:
+        - client_auth
+#  - name: pcap_api_platform_mtls
```
b1tamara, line 171: If it does not needed, remove these lines at all.

manifests/pcap-api.yml
a18e: I left these in intentionally for the followup where we properly enable platform TLS. But I can take them out if you prefer.

manifests/pcap-api.yml
a18e: Will be done as part of a followup.

manifests/pcap-api.yml
a18e: I left these in intentionally for the followup where we enable platform TLS

manifests/pcap-api.yml
```diff
@@ -60,14 +79,31 @@ instance_groups:
               - name: pcap-api-public-endpoint
                 registration_interval: 20s
                 port: 8080
+                #                tls_port: 8080
```
b1tamara, line 82: Remove if it is not needed
If port and tls_port are provided, then route_registrar takes the value of tls_port.

a18e: See https://github.com/cloudfoundry/pcap-release/pull/23/commits/f4553fbb7302ad917ff8e4b8413eb93c7c830fe6

manifests/pcap-api.yml
a18e: Removed, see https://github.com/cloudfoundry/pcap-release/pull/23/commits/f4553fbb7302ad917ff8e4b8413eb93c7c830fe6

jobs/pcap-api/templates/pcap-api.yml.erb
a18e: See https://github.com/cloudfoundry/pcap-release/pull/23/commits/f4553fbb7302ad917ff8e4b8413eb93c7c830fe6

maxmoehl (approved):

PR #25, opened 2023-05-11
ci: add semantic releases

domdom82:
- Add semantic-release to generate version and release notes
- Add "rc" concourse job to do a dry run of a release
- Add "shipit" job to build and publish a new release

### DIFF
--- a/.releaserc
+++ b/.releaserc
@@ -1,0 +1,26 @@
+{
+  "ci": false,
+  "branches": ["main", {"name": "alpha", "prerelease": true}, {"name": "beta", "prerelease": true}],
+  "plugins": [
+    ["@semantic-release/commit-analyzer", {
+      "preset": "angular",
+      "releaseRules": [
+        {"type": "dep", "release": "patch"}
+      ],
+      "parserOpts": {
+        "noteKeywords": ["BREAKING CHANGE", "BREAKING-CHANGE"]
+      }
+    }],
+    "@semantic-release/release-notes-generator",
+    ["@semantic-release/exec", {
+      "verifyReleaseCmd": "bosh -n create-release --final --version ${nextRelease.version} --tarball /tmp/pcap-${nextRelease.gitTag}.tgz",
+      "publishCmd": "./ci/scripts/publish.sh ${nextRelease.gitTag}",
+      "generateNotesCmd": "./ci/scripts/notes.sh ${nextRelease.gitTag} ${nextRelease.version}"
+    }],
+    ["@semantic-release/github", {
+      "assets": [
+        {"path": "/tmp/pcap-*.tgz"}
+      ]
+    }]
+  ]
+}
--- a/ci/Dockerfile
+++ b/ci/Dockerfile
@@ -5,6 +5,9 @@ ARG GINKGO_VERSION=latest
 ARG GOLANGCILINT_VERSION=latest
 RUN apt-get update && apt-get install -y libpcap-dev && rm -rf /var/lib/apt/lists/*
 
+RUN curl -fsSL https://deb.nodesource.com/setup_current.x | sudo -E bash - && \
+    apt-get install -y nodejs && rm -rf /var/lib/apt/lists/*
+
 RUN rm -rf /usr/local/go
 COPY --from=golang:1.20 /usr/local/go/ /usr/local/go/
 ENV GOPATH=/go PATH=${PATH}:/go/bin
--- a/ci/Dockerfile
+++ b/ci/Dockerfile
@@ -8,6 +8,9 @@ RUN apt-get update && apt-get install -y libpcap-dev && rm -rf /var/lib/apt/list
 RUN curl -fsSL https://deb.nodesource.com/setup_current.x | sudo -E bash - && \
     apt-get install -y nodejs && rm -rf /var/lib/apt/lists/*
 
+RUN npm install -g semantic-release && \
+    npm install -g @semantic-release/exec
+
 RUN rm -rf /usr/local/go
 COPY --from=golang:1.20 /usr/local/go/ /usr/local/go/
 ENV GOPATH=/go PATH=${PATH}:/go/bin
--- a/ci/pipeline.yml
+++ b/ci/pipeline.yml
@@ -248,4 +248,4 @@ resources:
     source:
       bucket:   pcap-release
       json_key: ((gcp.service_key))
-      regexp:   pcap-([0-9\.+-]+).tgz
+      regexp:   pcap-v[0-9a-z\.+-]+.tgz
--- a/ci/pipeline.yml
+++ b/ci/pipeline.yml
@@ -226,9 +226,9 @@ resources:
   - name: git
     type: git
     source:
-      uri: git@github.com:cloudfoundry/pcap-release.git
+      uri: https://github.com/cloudfoundry/pcap-release.git
       branch: main
-      private_key: ((github.private_key))
+      password: ((github.access_token))
 
   - name: git-pull-requests
     type: pull-request
--- a/ci/pipeline.yml
+++ b/ci/pipeline.yml
@@ -142,7 +142,7 @@ jobs:
             type: docker-image
             source:
               repository: cf-routing.common.repositories.cloud.sap/pcap-release-testflight
-              tag:        semantic
+              tag:        latest
               username:   ((docker.username))
               password:   ((docker.password))
           inputs:
@@ -186,7 +186,7 @@ jobs:
             type: docker-image
             source:
               repository: cf-routing.common.repositories.cloud.sap/pcap-release-testflight
-              tag:        semantic
+              tag:        latest
               username:   ((docker.username))
               password:   ((docker.password))
           inputs:
--- a/ci/pipeline.yml
+++ b/ci/pipeline.yml
@@ -4,6 +4,8 @@ groups:
     jobs:
       - unit-tests
       - unit-tests-pr
+      - shipit
+      - rc
 
 jobs:
   - name: unit-tests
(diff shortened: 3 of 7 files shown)


.releaserc
```diff
@@ -0,0 +1,26 @@
+{
+  "ci": false,
+  "branches": ["main", {"name": "alpha", "prerelease": true}, {"name": "beta", "prerelease": true}],
+  "plugins": [
+    ["@semantic-release/commit-analyzer", {
+      "preset": "angular",
+      "releaseRules": [
+        {"type": "dep", "release": "patch"}
```
peanball, line 8: Are these exclusive, or in addition to preset `angular`?

.releaserc
```diff
@@ -0,0 +1,26 @@
...
+      ],
+      "parserOpts": {
+        "noteKeywords": ["BREAKING CHANGE", "BREAKING-CHANGE"]
```
peanball, line 11: I think you mentioned in our internal ticket to use `breaking-change:` as well as trailer, or is that case insensitive?

ci/Dockerfile
```diff
@@ -5,6 +5,9 @@ ARG GINKGO_VERSION=latest
 ARG GOLANGCILINT_VERSION=latest
 RUN apt-get update && apt-get install -y libpcap-dev && rm -rf /var/lib/apt/lists/*
 
+RUN curl -fsSL https://deb.nodesource.com/setup_current.x | sudo -E bash - && \
```
peanball, line 8: Should we consider pulling in the data from `node:20.x`/ `node:current` or similar?

```Dockerfile
ARG NODE_VERSION=current
FROM node:${NODE_VERSION} as nodejs
COPY --from=nodejs /usr/bin/node
```
(along the lines of this, not tested, just as idea)

peanball: See also https://github.com/moby/moby/issues/45419#issuecomment-1525256205

ci/pipeline.yml
```diff
@@ -126,6 +128,84 @@ jobs:
...
+      - in_parallel:
+          - { get: git,     passed: [rc] }
+      - task: release
+        config:
+          platform: linux
+          image_resource:
+            type: docker-image
+            source:
+              repository: cf-routing.common.repositories.cloud.sap/pcap-release-testflight
+              tag:        semantic
```
peanball, line 145: Will this also still change?
Eventually we could have this on `latest` or some other commonly used tag name.

ci/pipeline.yml
```diff
@@ -166,3 +242,10 @@ resources:
     type: slack-notification
     source:
       url: ((slack.pcap_release_webhook))
+
+  - name: blobstore
+    type: gcs
+    source:
+      bucket:   pcap-release
+      json_key: ((gcp.service_key))
+      regexp:   pcap-([0-9\.+-]+).tgz
```
peanball, line 251: this would disallow `alpha`, etc, in case we have those in the version names anywhere.

ci/scripts/shipit
```diff
@@ -0,0 +1,52 @@
...
+: "${GCP_SERVICE_KEY:?required}" # The GCP service key for accessing the blobstore, written to a temporary private.yml.
+: "${GITHUB_TOKEN:?required}" # The github access token
+: "${DRY_RUN:=false}" # Whether to do a dry run or a real release
+
+###############################################################
+
+CONCOURSE_ROOT="$(pwd)"
+export CONCOURSE_ROOT
+
+DRY_RUN_ARG="--dry-run=$DRY_RUN"
```
peanball, line 25: you _could_ init that to a default, e.g. always do dry run if not explicitly disabled, or the other way around.

e.g.
```suggestion
DRY_RUN=${DRY_RUN:-false}
DRY_RUN_ARG="--dry-run=$DRY_RUN"
```

ci/scripts/shipit
```diff
@@ -0,0 +1,52 @@
...
+${PADDED_GCP_SERVICE_KEY}
+YAML
+
+header "Pulling in any git submodules..."
+git submodule update --init --recursive --force
+cd -
+
+cd "${REPO_ROOT}"
+header "Installing dependencies for semantic-release..."
+npm install @semantic-release/exec -D
```
peanball, line 49: Could we maybe even put those into our base image instead of having to pull them every time?

ci/scripts/notes.sh
```diff
@@ -0,0 +1,28 @@
+#!/bin/bash
+
+set -e
+
+RELEASE_GIT_TAG="$1"
+VERSION="$2"
```
peanball, line 6:
```suggestion
RELEASE_GIT_TAG="${1:?Release Git Tag is mandatory}"
VERSION="${2:?Version defining the target release version is mandatory}"
```

ci/pipeline.yml
```diff
@@ -126,6 +128,84 @@ jobs:
...
+            args: []
+          params:
+            RELEASE_NAME: pcap-release
+            REPO_ROOT:    git
+            RELEASE_ROOT: gh
+            BRANCH:        main
+            DRY_RUN: true
+            GITHUB_OWNER:  cloudfoundry
+            GITHUB_TOKEN: ((github.access_token))
+            GCP_SERVICE_KEY: ((gcp.service_key))
```
peanball, line 207: The indentation / alignment is a bit all over the place in this file, particularly here. Maybe we can align or bite the bullet and IDE auto format?

ci/Dockerfile
```diff
@@ -5,6 +5,9 @@ ARG GINKGO_VERSION=latest
...
+    apt-get install -y nodejs && rm -rf /var/lib/apt/lists/*
```
peanball, line 9: Consider installing semantic-release here right away, see also comment below in `shipit`, where it is installed every time.

peanball (changes requested):
Some points to consider, primarily related to the Dockerfile and its content.

peanball (changes requested):
Some points to consider, primarily related to the Dockerfile and its content.

peanball (changes requested):
Some points to consider, primarily related to the Dockerfile and its content.

.releaserc
domdom82: in addition. so angular + whatever rules we decide.

.releaserc
domdom82: after some discussion we decided to use the upper-case version to put emphasize on the importance of the breaking change. Also the [spec](https://www.conventionalcommits.org/en/v1.0.0/#specification) explicitly states that 
>The units of information that make up Conventional Commits MUST NOT be treated as case sensitive by implementors, with the exception of BREAKING CHANGE which MUST be uppercase.

ci/Dockerfile
domdom82: we don't just need node but npx to run semantic-release. So I'm not sure this way would work or what else would be needed to be copied. For go it's easy because you only have one binary.
So unless I'd have to go a different way, I'd say we keep it as it is now.

This is also the official way to install nodejs on ubuntu, see https://nodejs.org/en/download/package-manager#debian-and-ubuntu-based-linux-distributions

ci/pipeline.yml
domdom82: yes, we will tag this as "latest" once we are sure we didn't break anything

ci/scripts/shipit
domdom82: we do that. see line 18

ci/pipeline.yml
domdom82: not true. we have successfully made a "semantic-release" version, e.g. [pcap-v1.0.0-semantic-releases.1.tgz](https://console.cloud.google.com/storage/browser/_details/pcap-release/pcap-v1.0.0-semantic-releases.1.tgz?project=app-runtime-platform-wg)

This could also be an "alpha" version, e.g. pcap-v1.0.0-alpha.1.tgz as per the [naming convention](https://github.com/semantic-release/semantic-release/blob/master/docs/support/FAQ.md#can-i-set-the-initial-release-version-of-my-package-to-001), where pre-releases are added by their branch name into the version name.

ci/pipeline.yml
domdom82: let's decide on a formatter and use it everywhere. takes tomato vs tomato out of the picture.

ci/pipeline.yml
domdom82: as discussed: the regex is probably ignored somehow

ci/scripts/shipit
domdom82: couldn't get it to work with a global install. will give it another shot.

ci/pipeline.yml
domdom82: image tagged to latest.

ci/Dockerfile
domdom82: I did meanwhile manage to add everything to the dockerfile.

ci/scripts/shipit
domdom82: does work now.

ci/Dockerfile
```diff
@@ -5,6 +5,12 @@ ARG GINKGO_VERSION=latest
...
+
+RUN npm install -g semantic-release && \
+    npm install -g @semantic-release/exec
```
peanball, line 12: This could be in the line above. NPM should be able to install different packages at once.

domdom82: i'd rather have each package on its own line for clarity. yes we can install two packaged in one command though.

ci/pipeline.yml
domdom82: fixed regex to match alphanumeric values in the version, too.

peanball (approved):

PR #29, opened 2023-06-05
acceptance-tests

peanball:
- Prepared test suite environment
- Added one basic integration test
- Added several fixes of bugs that surfaced during tests

Successful integration test run: https://concourse.cfi.sapcloud.io/teams/main/pipelines/pcap-release-dev/jobs/acceptance-tests/builds/89


acceptance-tests/basic_pcap_test.go
```diff
@@ -0,0 +1,68 @@
...
+
+		info, _ := deployPcap(
+			baseManifestVars{
+				deploymentName: deploymentNameForTestNode(),
+			},
+			[]string{},
+			map[string]interface{}{},
+			true,
+		)
+
```
peanball, line 25: We could consider moving this into the `Describe` block above, or a `Context`.

This is of course unless the different tests within a single test file would be executed on different nodes with different `deploymentNameForTestNode()`?

acceptance-tests/basic_pcap_test.go
```diff
@@ -0,0 +1,68 @@
...
+		}
+
+		By("Stopping capture after curl has finished")
+		sessionPcap.Interrupt()
+		Eventually(sessionPcap, time.Minute, time.Second).Should(gexec.Exit())
+
+		By("Checking that the capture has produced a valid pcap file")
+		pcapFileStat, err := os.Stat(pcapFile)
+		Expect(err).NotTo(HaveOccurred())
+		Expect(pcapFileStat.Size()).To(BeNumerically(">", 24)) // 24 bytes == pcap header only
```
peanball, line 66: Please add explanations for expectations and their failures via the `optionalDescription` field
```suggestion
		Expect(pcapFileStat.Size()).To(BeNumerically(">", 24, "Only the pcap header is contained in the capture file. Expected more data.")) 
```

acceptance-tests/acceptance_tests_suite_test.go
```diff
@@ -0,0 +1,50 @@
...
+	RegisterFailHandler(Fail)
+	RunSpecs(t, "AcceptanceTests Suite")
+}
+
+var _ = SynchronizedBeforeSuite(func() []byte {
+	// Load config once, and pass to other
+	// threads as JSON-encoded byte array
+	var err error
+	config, err = loadConfig()
+	Expect(err).NotTo(HaveOccurred())
```
peanball, line 21: Please add explanations for expectations and their failures via the `optionalDescription` field.

This comment applies to pretty much every `Expect()`.

```suggestion
	Expect(err).NotTo(HaveOccurred(), "loading common config")
```

peanball (commented):
I would really appreciate if we added descriptions with the logical expectation for each of the `Expect()` statements.

Failures are then much more easily identified by looking at the logs, and it helps see, verify or question the test expectations.

acceptance-tests/bosh_helpers.go
```diff
@@ -0,0 +1,353 @@
...
+}
+
+func writeLog(s string) {
+	ginkgoConfig, _ := GinkgoConfiguration()
+	for _, line := range strings.Split(s, "\n") {
+		fmt.Printf("node %d/%d: %s\n", ginkgoConfig.ParallelProcess, ginkgoConfig.ParallelTotal, line)
+	}
+}
+
+func downloadFile(info pcapInfo, remotePath string, localFile *os.File, permissions os.FileMode) error {
```
peanball, line 351: This function would benefit significantly from godoc, as the term 'download' is logically overloaded and in this case actually means "SCP".

```suggestion
// downloadFile retrieves the file from remotePath via SSH and stores it in localFile with permissions. info contains the remote's credentials and IP address.
// See copyFileFromRemote
func downloadFile(info pcapInfo, remotePath string, localFile *os.File, permissions os.FileMode) error {
```

src/pcap/cmd/pcap-bosh-cli/main.go
```diff
@@ -200,8 +202,11 @@ func checkOutputFile(file string, overwrite bool) error {
 	}
 	// File doesn't exist, check if path is valid
 	fileInfo, err := os.Stat(filepath.Dir(file))
-	if err != nil || !fileInfo.IsDir() {
-		return fmt.Errorf("cannot write file %s. %s does not exist", file, fileInfo.Name())
+	if err != nil {
+		if fileInfo != nil {
```
b1tamara, line 206: Do we really need this additional check? It might be enough to return on error something like: 
`fmt.Errorf("cannot write file %s due to error: %w", file, err)`

acceptance-tests/remote.go
```diff
@@ -0,0 +1,236 @@
...
+	"log"
+	"net"
+	"os"
+	"time"
+
+	"github.com/bramvdbogaerde/go-scp"
+	"golang.org/x/crypto/ssh"
+)
+
+// runs command on remote machine
```
peanball, line 17: please adapt to a proper godoc comment.

Applies to all functions in this file.

acceptance-tests/remote.go
```diff
@@ -0,0 +1,236 @@
...
+
+	scpClient := scp.NewClient(fmt.Sprintf("%s:22", addr), clientConfig)
+	if err := scpClient.Connect(); err != nil {
+		return err
+	}
+
+	return scpClient.CopyFile(context.Background(), fileReader, remotePath, permissions)
+}
+
+func copyFileFromRemote(user string, addr string, privateKey string, remotePath string, localFile *os.File, permissions os.FileMode) error {
```
peanball, line 52: This is wrapped with `downloadFile` in the bosh_helper.go file.

Is there a scenario where we would issue a `copyFileFromRemote()` without having a `pcapInfo`?

ci/scripts/start-bosh.sh
```diff
@@ -0,0 +1,230 @@
+#!/usr/bin/env bash
```
peanball, line 1: Do we need our own start-bosh file or would the one from the docker image `bosh/docker-cpi:main` also work?

src/pcap/cmd/pcap-bosh-cli/main.go
```diff
@@ -393,31 +417,31 @@ func (e *Environment) connect() error {
...
-	}
-
-	transport := http.DefaultTransport.(*http.Transport).Clone()
-	transport.TLSClientConfig.RootCAs = boshCA
-
-	e.client = &http.Client{
-		Transport: transport,
-	}
+	} else {
+		logger.Info("using TLS-encrypted connection to bosh-director", zap.String("bosh-director-url", e.DirectorURL.String()))
```
b1tamara, line 429: Why do you want to introduce else block to the condition? The further code is more readable without else block.

src/pcap/cmd/pcap-bosh-cli/main.go
```diff
@@ -38,9 +41,8 @@ type options struct {
 	PcapAPIURL         string   `short:"u" long:"pcap-api-url" description:"The URL of the PCAP API, e.g. pcap.cf.$LANDSCAPE_DOMAIN" env:"PCAP_API" required:"true"`
 	Filter             string   `short:"f" long:"filter" description:"Allows to provide a filter expression in pcap filter format." required:"false"`
 	Interface          string   `short:"i" long:"interface" description:"Specifies the network interface to listen on." default:"eth0" required:"false"`
-	Type               string   `short:"t" long:"type" description:"Specifies the type of process to capture for the app." default:"web" required:"false"`
 	BoshConfigFilename string   `short:"c" long:"bosh-config" description:"Path to the BOSH config file, used for the UAA Token" default:"${HOME}/.bosh/config" required:"false"`
-	BoshEnvironment    string   `short:"e" long:"bosh-environment" description:"The BOSH environment to use for retrieving the BOSH UAA token from the BOSH config file" default:"bosh" required:"false"`
+	BoshEnvironment    string   `short:"e" long:"bosh-environment" description:"The BOSH environment to use for retrieving the BOSH UAA token from the BOSH config file" env:"BOSH_ENVIRONMENT" required:"false"`
```
peanball, line 45: Are `env:` and `default:` mutually exclusive? If not, it would be better to support both.

I rarely set `BOSH_ENVIRONMENT` and instead call `bosh -e the_env_that_i_want -d ...`. Not having to worry about it and defaulting to `bosh`, even when the env variable is not set would still be nice.

If it's mutually exclusive, I agree that the env variable has higher precedence.

src/pcap/cmd/pcap-bosh-cli/main.go
```diff
@@ -382,8 +390,24 @@ func (e *Environment) UpdateTokens() error {
...
+
+	err = e.fetchUAAURL()
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// setup configures the BOSH http client for a given environment.
```
peanball, line 407: in the godoc please explain the URL autosensing, especially with the new functionality of determining HTTP vs. HTTPS based on the existence of a CA certificate.

ci/scripts/shell
```diff
@@ -0,0 +1,43 @@
+#!/bin/bash
```
b1tamara, line 1: Could you maybe describe the purpose of the script?

manifests/pcap-acceptance-tests.yml
```diff
@@ -0,0 +1,106 @@
...
+        - client_auth
+        - server_auth
+  - name: pcap_agent_mtls
+    options:
+      alternative_names:
+        - pcap-agent.service.cf.internal
+      ca: pcap_ca
+      common_name: pcap_agent_mtls
+      extended_key_usage:
+        - client_auth
```
b1tamara, line 104: minor thing: it might be not needed for agent certs.

acceptance-tests/config.go
```diff
@@ -0,0 +1,141 @@
...
+		fmt.Sprintf("BOSH_ENVIRONMENT=%s", config.BoshEnvironment),
+		fmt.Sprintf("HOME=%s", config.HomePath),
+		fmt.Sprintf("BOSH_DEPLOYMENT=%s", boshDeployment),
+		"BOSH_NON_INTERACTIVE=true",
+	}
+
+	return cmd
+}
+
+func (config *Config) boshInteractiveCmd(args ...string) *exec.Cmd {
```
b1tamara, line 121: We can have a boshBasicCmd like 
```
func (config *Config) boshBasicCmd(args ...string) *exec.Cmd {
	cmd := exec.Command(config.BoshPath, append([]string{"--tty", "--no-color"}, args...)...)
	cmd.Env = []string{
		fmt.Sprintf("BOSH_DIRECTOR_IP=%s", config.BoshDirectorAPI),
		fmt.Sprintf("BOSH_DIRECTOR_CA=%s", config.BoshDirectorCA),
		fmt.Sprintf("BOSH_DIRECTOR_CERT=%s", config.BoshDirectorCert),
		fmt.Sprintf("BOSH_ENVIRONMENT=%s", config.BoshEnvironment),
		fmt.Sprintf("HOME=%s", config.HomePath),
	}

	return cmd
}
```

And then reuse the function : 
```
func (config *Config) boshCmd(boshDeployment string, args ...string) *exec.Cmd {
	cmd := config.boshBasicCmd(args)
	cmd.Env = append(cmd.Env, 
		fmt.Sprintf("BOSH_CLIENT=%s", config.BoshClient),
		fmt.Sprintf("BOSH_CLIENT_SECRET=%s", config.BoshClientSecret),
		fmt.Sprintf("BOSH_DEPLOYMENT=%s", boshDeployment),
		"BOSH_NON_INTERACTIVE=true")

	return cmd
}

func (config *Config) boshInteractiveCmd(args ...string) *exec.Cmd {
	return config.boshBasicCmd(args)
}
```

acceptance-tests/basic_pcap_test.go
```diff
@@ -0,0 +1,68 @@
...
+	. "github.com/onsi/gomega"
+	"github.com/onsi/gomega/gexec"
+	"net/http"
+	"os"
+	"os/exec"
+	"time"
+)
+
+var _ = Describe("Pcap Deployment", func() {
+	It("Deploys successfully", func() {
```
b1tamara, line 15: We can have two test cases for that. 
The first one - "Pcap deploys successfully" which ensures only that the release can be deployed 
The second one - "Traffic captured successfully"

acceptance-tests/basic_pcap_test.go
domdom82: >This is of course unless the different tests within a single test file would be executed on different nodes with different deploymentNameForTestNode()?

Yes, they are. All `It`s are randomized and run in parallel. (we only have one here, but you get the point :)
Each test means a different deployment name and we have several run in parallel so each needs a different name to avoid clashes.

acceptance-tests/basic_pcap_test.go
domdom82: The "can be deployed" part is tested implicitly with the `SynchronizedBeforeSuite` function. So I thought it makes little sense to then again test what we just tested and I added a little extra, also to keep the runtime low.

But sure, in the future we can add more detailed tests with different scenarios.

acceptance-tests/basic_pcap_test.go
```diff
@@ -0,0 +1,68 @@
...
+		)
+
+		By("Logging on to BOSH director to get a refresh token")
+		login(config.BoshClient, config.BoshClientSecret)
+
+		pcapBoshCliFile, err := os.CreateTemp("", "pcap-bosh-cli-*")
+		Expect(err).NotTo(HaveOccurred())
+		pcapBoshCli := pcapBoshCliFile.Name()
+
+		By("Downloading remote pcap-bosh-cli-linux-amd64 to " + pcapBoshCli)
```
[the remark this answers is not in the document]
domdom82: OK, unfortunate, btw. I liked the "download CLI from API" feature from the PoC...

acceptance-tests/basic_pcap_test.go
```diff
@@ -0,0 +1,68 @@
...
+		Expect(err).NotTo(HaveOccurred())
+
+		pcapFile := fmt.Sprintf("%s-capture.pcap", pcapBoshCli)
+		By("Starting capture of traffic on pcap-agent instance to file " + pcapFile)
+		cmdPcap := exec.Command(
+			pcapBoshCli,
+			"-d", deploymentNameForTestNode(),
+			"-g", "pcap-agent",
+			"-o", pcapFile,
+			"-u", fmt.Sprintf("http://%s:8080/", info.PcapAPIPublicIP),
```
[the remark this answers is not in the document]
domdom82: yes. I planned to do that for future tests where we also test the TLS config that runs on a different port. I can add a TODO for now.

acceptance-tests/basic_pcap_test.go
```diff
@@ -0,0 +1,68 @@
...
+			"-v")
+		sessionPcap, err := gexec.Start(cmdPcap, GinkgoWriter, GinkgoWriter)
+		Expect(err).NotTo(HaveOccurred())
+
+		By("Curling apache on pcap-agent instance to produce some traffic")
```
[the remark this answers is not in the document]
domdom82: oh c'mon now...

acceptance-tests/basic_pcap_test.go
domdom82: fair enough.

acceptance-tests/bosh_helpers.go
```diff
@@ -0,0 +1,353 @@
...
+
+	By("Checking /var/vcap/jobs/pcap-agent/config/pcap-agent.yml")
+	pcapAgentConfig, _, err := runOnRemote(pcapInfo.SSHUser, pcapInfo.PcapAgentPublicIP, pcapInfo.SSHPrivateKey, "cat /var/vcap/jobs/pcap-agent/config/pcap-agent.yml")
+	Expect(err).NotTo(HaveOccurred())
+	writeLog("---------- PcapAPI Config ----------")
+	writeLog(pcapAgentConfig)
+	writeLog("------------------------------------")
+}
+
+// Takes bosh deployment name, ops files and vars.
```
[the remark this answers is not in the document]
domdom82: copy&paste from HAproxy :) valid point though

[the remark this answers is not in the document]
domdom82: Yes. A lot of is basically copied over from HAProxy, though. I'd rather see a godoc for a possible shared module that each of the projects may use, in a separate repository.

acceptance-tests/bosh_helpers.go
domdom82: Yes. It is a sibling of [uploadFile](https://github.com/cloudfoundry/haproxy-boshrelease/blob/master/acceptance-tests/https_ext_crt_list_test.go#L420) in HAproxy. There it's even worse, the function is only defined in one test and with no docs at all. I'd say we will have a follow-up discussion on a shared module for these two releases.

acceptance-tests/config.go
domdom82: yes, but it would change the meaning. the point was to have one interactive command that requires user input and one non-interactive command that expects user creds as variables. Your `boshBasicCmd` is the same as `boshInteractiveCmd`.

So if at all, you would want to have `boshInteractiveCmd` be the "basic" one and let `boshCmd` add to it, but that also feels kinda off.

acceptance-tests/remote.go
domdom82: yes, there is. `downloadFile` currently implies `pcap-api` as the target. If you want to copy stuff from somewhere else you would use `copyFileFromRemote`. Granted, `downloadFile` is a convenience function.

acceptance-tests/remote.go
domdom82: ok, will do. Can you open an issue with HAProxy bosh release to do the same there? Most of this code was copied over.

ci/scripts/shell
domdom82: Yes. The script was copied from HAproxy where there is a README for acceptance tests that describes it. I could also add a read me to pcap.

manifests/pcap-acceptance-tests.yml
domdom82: true. I'm removing it.

src/pcap/cmd/pcap-bosh-cli/main.go
domdom82: I've had both and it wouldn't read my env var. So I think they are mutually exclusive. Anyway, we can't make any default assumptions on the name of a BOSH env. The BOSH cli also does not do this, it tries to read this env var or expects an explicit parameter, otherwise throw an error. I've tried to align on that behavior here.

src/pcap/cmd/pcap-bosh-cli/main.go
domdom82: I wanted to retain the original behavior as much as possible, where the error message would print the directory that's missing. Before my change the code assumed `fileInfo` to never be nil which was clearly wrong.

src/pcap/cmd/pcap-bosh-cli/main.go
domdom82: we don't. if the URL contains no scheme, we assume HTTPS as with the pcap API. no other change was made, because there is no way to talk to BOSH director without TLS anyway.

domdom82:
Green integration test after code review changes applied: https://concourse.cfi.sapcloud.io/teams/main/pipelines/pcap-release-dev/jobs/acceptance-tests/builds/92

acceptance-tests/config.go
b1tamara: I did not want to rename the function `boshInteractiveCmd`, as you are right it will change the meaning of the function. But I wanted to have the same coding on one place.

ci/scripts/shell
b1tamara: I cannot find in HAProxy acceptance-tests README or in ci README about this script

b1tamara: It is a good point to add a README file which contains how to run the scripts and tests locally.

acceptance-tests/basic_pcap_test.go
peanball: But with `BeforeEach` you'd still have it in the context of the same "thread", i.e.
1. `BeforeEach`
2. `It`
3. (`AfterEach`)

acceptance-tests/basic_pcap_test.go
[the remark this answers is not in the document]
peanball:
```suggestion
 		By("Issuing a request to apache on pcap-agent instance to produce some traffic")
```

acceptance-tests/bosh_helpers.go
peanball: We have a follow-up for unifying this. But I also don't want to accept new technical dept into pcap because we just copied from HAProxy. What we build here can be the base of the unified version that we'd also use in HAProxy.

acceptance-tests/remote.go
peanball: done. https://github.com/cloudfoundry/haproxy-boshrelease/issues/482

ci/scripts/shell
peanball: Again, I would like to use the opportunity of us touching all these scripts (even if they're copied from HAProxy) to improve them. A lot of this improvement is documentation. We cannot introduce new technical dept in pcap-release though.

peanball (commented):
I would really like to avoid introducing new technical dept, even if we reuse test code / scripts from HAProxy. Those are not or barely documented. This is the ideal opportunity to improve.

We can / will still unify across haproxy and pcap releases, but if we have a better base to jump off from that would be beneficial for both projects.

acceptance-tests/basic_pcap_test.go
b1tamara: Hm... you are right. But maybe it makes sense to make it more explicitly.  Especially for the test explicit case if deploy will not work. Then it will be very fast clear what use case is broken.

acceptance-tests/basic_pcap_test.go
domdom82: yes, makes sense when there is more than one test.

ci/scripts/shell
domdom82: I've added acceptance-tests/README.md and a comment to this script that explains how it works and what the intention is.

acceptance-tests/README.md
```diff
@@ -0,0 +1,101 @@
+# Acceptance Tests
+
+## Requirements
+
+* Docker installed locally
+* A matching Bionic stemcell tgz downloaded to `ci/scripts/stemcell-bionic`
+    * Get it from https://bosh.io/stemcells/bosh-warden-boshlite-ubuntu-bionic-go_agent
```
b1tamara, line 7: bionic or jammy?

domdom82: good catch. correct it to jammy.

domdom82 (approved):

PR #30, opened 2023-06-06
fix: request and verify client certificates if ca is configured

maxmoehl:
If one or more certificates are configured for client authentication we now automatically request and verify certificates from the client.


peanball (approved):

maxmoehl:
It might be that we want to make this configurable eventually but for now I don't see the need.

PR #41, opened 2023-06-26
Add golang-release as vendored bosh package

peanball:
Adds the vendored bosh package with adaptations to packaging scripts.

This is deliberately outdated so the autobumper can be tested.


packages/pcap-api/spec
```diff
@@ -4,6 +4,5 @@ name: pcap-api
 dependencies: []
```
maxmoehl, line 4: This should now depend on the golang package.

maxmoehl (changes requested):

packages/pcap-api/spec
maxmoehl: Same goes for the agent.

maxmoehl (dismissed):

peanball:
Verified working on a dev landscape

maxmoehl (approved):

PR #45, opened 2023-06-27
Add our convention to dependabot for commit prefixes

peanball:
We use:
* `dep` for dependency updates in the resulting build
* `ci` for dependencies in the build infrastructure that don't directly have an impact no the build output

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -7,6 +7,10 @@ updates:
     day: "monday"
     time: "09:00"
     timezone: "Europe/Berlin"
+  commit-message:
+    prefix: "ci"
+  labels:
+    - run-ci
 - package-ecosystem: gomod
   directory: /src/pcap
   schedule:
@@ -14,3 +18,7 @@ updates:
     day: "monday"
     time: "09:00"
     timezone: "Europe/Berlin"
+  commit-message:
+    prefix: "dep"
+  labels:
+    - run-ci


.github/dependabot.yml
```diff
@@ -7,10 +7,23 @@ updates:
     day: "monday"
     time: "09:00"
     timezone: "Europe/Berlin"
+  commit-message:
+    prefix: "ci"
+  labels:
+    - run-ci
```
maxmoehl, line 13: Are we sure that we want to always run those PRs automatically? I thought we had some discussions about this in the past and decided against it.

.github/dependabot.yml
```diff
@@ -7,10 +7,23 @@ updates:
...
 - package-ecosystem: gomod
   directory: /src/pcap
   schedule:
     interval: weekly
     day: "monday"
     time: "09:00"
     timezone: "Europe/Berlin"
+  commit-message:
+    prefix: "dep"
+    prefix-development: "ci"
```
maxmoehl, line 25: This has no effect for the `gomod` package-ecosystem.

.github/dependabot.yml
```diff
@@ -7,10 +7,22 @@ updates:
     day: "monday"
     time: "09:00"
     timezone: "Europe/Berlin"
+  commit-message:
+    prefix: "ci"
+  labels:
+    - run-ci
+  reviewers:
+    - "cloudfoundry/wg-app-runtime-platform-networking-extensions-approvers"
```
maxmoehl, line 15: Since we have code-owners set up this shouldn't be needed?

peanball: yes. removed.

maxmoehl (approved):

PR #51, opened 2023-06-27
fix: Create only one PR for one dependency update

b1tamara:


ci/scripts/autobump-dependencies.py
```diff
@@ -309,37 +320,34 @@ def main() -> None:
...
+        for package in dependency.packages:
+            current_version = dependency.check_current_version(package)
+            latest_release = dependency.latest_release
+            latest_version = latest_release.version
+
+            if latest_version <= current_version:
+                print(f"[{dependency.name}] in {package} already on the latest version: {latest_version} " f"(pinned: {dependency.pinned_version}.*)")
+                continue
+
+            if dependency.open_pr_exists():
```
peanball, line 337: I think the PR should be checked before iterating through the packages, as we want one PR per dependency, right?

peanball (changes requested):
I think the "PR for this dependency already exists" check can be pulled into the outer loop.

ci/scripts/autobump-dependencies.py
peanball: Nevermind. As discussed, this approach is fine.

peanball (approved):

PR #74, opened 2023-07-05
Disable acceptance-tests linting for now.

domdom82:
We will follow-up as discussed.


peanball (approved):
Approved with the condition that we follow up enabling linters and addressing existing linter issues in the current code.

Some of the linter used in `src/pcap` would not be applicable to test code.

PR #76, opened 2023-07-06
Enforce commits are following our conventions via github actions

raghunathd8:
Adds a commitlint check job that ensures the following rules:

1. conventional commits
2. supported commit types: `feat`, `fix`, `dep`, `doc`, `ci`, `refactor`, `test`
3. Breaking changes can be indicated with a `!` before `:`, e.g. `feat!: new breaking feature`.  
   Breaking changes _must_ have a `breaking-change:` or `breaking change:` trailer. This is case in-sensitive.

The commitlint configuration ensures those settings. The GitHub Actions job executes them and fails the PR validation if any of the commits in a PR are not compliant. Those can then be amended as needed by the PR author.

### DIFF
--- a/.github/workflows/commitlint.yml
+++ b/.github/workflows/commitlint.yml
@@ -1,0 +1,21 @@
+name: CI
+
+on:
+  pull_request:
+    branches:
+      - main
+jobs:
+  commitlint:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+      - name: Setup Node.js
+        uses: actions/setup-node@v2
+        with:
+          node-version: '18.x'
+      - name: Install commitlint
+        run: npm install -g commitlint
+      - name: Validate PR commits with @commitlint/cli
+        run: npx commitlint --from ${{ github.event.pull_request.head.sha }}~${{ github.event.pull_request.commits }} --to ${{ github.event.pull_request.head.sha }} --verbose
--- a/commitlint.config.js
+++ b/commitlint.config.js
@@ -1,0 +1,71 @@
+module.exports = {
+    rules: {
+        // rules are defined as [ severity (2=error), "always"|"never", argument? ]. See
+        // https://commitlint.js.org/#/reference-rules?id=rules
+
+        // log-commit-info is used for debugging. It outputs the dict passed to a plugin.
+        // 'log-commit-info': [2, 'always'],
+        // enforce that commits with a breaking change indicator ("!") must also contain an explanation of the breaking change.
+        'breaking-change-indicator': [2, 'always'],
+        // enforce that only the types below are allowed
+        'type-enum': [2, 'always',
+            [
+                'feat',
+                'fix',
+                'dep',
+                'ci',
+                'refactor',
+                'doc',
+                'test',
+            ]
+        ],
+        // enforce that a type is mandatory
+        'type-empty': [2, 'never'],
+        // enforce the maximum subject length (without type and scope)
+        'subject-max-length': [2, 'always', 75],
+        // enforce the maximum scope length
+        'scope-max-length': [2, 'always', 25],
+    },
+    parserPreset: {
+        parserOpts: {
+            // add matching groups for `!` before and after scope. Before scope is invalid, after scope is correct
+            // and enforces the "breaking-change" header (see below).
+            headerPattern: /^(\w*)(?:(!?)\(([\w\$\.\-\* ]*)\))?(!?)\: (.*)$/,
+            headerCorrespondence: ["type", "wrong_breaking", "scope", "breaking", "subject"],
+        },
+    },
+    plugins: [
+        {
+            rules: {
+                // log-commit-info is used for debugging. It outputs the dict passed to a plugin.
+                // 'log-commit-info': (input) => {
+                //     console.log(input)
+                //     return [true, "see above"]
+                // },
+                'breaking-change-indicator': ({wrong_breaking, breaking, scope, body, footer}) => {
+                    if (wrong_breaking) {
+                        return [
+                            false, "The breaking change indicator `!` can only be used right before `:`, e.g. `feat(scope)!:`."
+                        ]
+                    }
+
+                    const rxBreakingChange = /^breaking(\s|-)change: .+/img
+
+                    // The marker "breaking change:" is treated as footer and "note" by commitlint, and not considered part of the body.
+                    // When a breaking change indicator ("!") is there, we need to check body and footer to cover both variants (with and without "-").
+
+                    // `commitOk` indicates, if the commit passes the rules.
+                    const commitOk =
+                        !breaking // if there is no "!" in the commit, we don't enforce a footer
+                        || body?.match(rxBreakingChange) // if there is a "!", and a body exists, it must match the breaking change pattern
+                        || footer?.match(rxBreakingChange) // if there is a "!", and a footer exists, it must match the breaking change pattern
+
+                    return [
+                        commitOk,
+                        'For a breaking change (indicated with `!` after the type), the commit must mention a trailer `breaking-change:` or `breaking change:` (case in-sensitive)'
+                    ];
+                },
+            },
+        },
+    ],
+};


.github/workflows/commitlint.yml
```diff
@@ -0,0 +1,30 @@
+name: CI
+
+on: [pull_request]
```
maxmoehl, line 3: The format is slightly different, see [this example](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-including-branches).
```suggestion
on:
  pull_request:
    branches:
      - main
```

peanball:
The trailer is in the wrong format in the commit, should be:
```
Co-authored-by: Soha Alboghdady <soha.alboghdady@sap.com>
```

`<` vs. `(`.

maxmoehl:
@plowin if you want to run the CI you have to approve the PR ;)

.github/workflows/commitlint.yml
```diff
@@ -0,0 +1,32 @@
...
+        run: |
+          node --version
+          npm --version
+          npx commitlint --version
+      - name: Install commitlint
+        run: |
+          npm install -g @commitlint/cli @commitlint/config-conventional
+      - name: Configure commitlint
+        run: |
+          echo "module.exports = {extends: ['@commitlint/config-conventional']}" > commitlint.config.js
```
peanball, line 29: As noted in our team discussion, this should ideally be either in our regular image that we also use in concourse, or a separate minimal image that contains commitlint. 

Fetching and installing these packages every time would only be reasonable if those were dependencies that are vital to the build or resulting output of the build, so we could ensure that the build would still work. 

With commit linting we only care that it's actually checking the commit message. If that were to break with a newer version, we could still stay on the old and adapt in time. It would not be a reproducibility blocker.

peanball (dismissed):

.github/workflows/commitlint.yml
peanball: Agreed on using a node base image and eventually switching to a prepared commitlint docker image, when upstream provides one.

peanball:
@Soha-Albaghdady it seems that GitHub actions is picky about what it supports as base image on public runners.

https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources

See also https://stackoverflow.com/questions/70959954/error-waiting-for-a-runner-to-pick-up-this-job-using-github-actions

So maybe we need to change back to `ubuntu-latest` and install npm.

There is an action for that though, that allows you to select the NPM version. Action resources should also be cached and should be the preferred way over downloading via script on the fly.

raghunathd8:
> https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources

This is exactly what I was referring to, github have selected specific runners that they support and encourage to use.

peanball:
one final bit for the commit, as this is a change on the overall CI setup, I would propose the prefix `ci: `.

peanball (approved):

.github/workflows/commitlint.yml
```diff
@@ -0,0 +1,31 @@
...
```
peanball, line 28: There is so far no commitlint configuration, right?

Which types does it allow and which ones will it not accept? Or will it just accept anything that looks like a conventional commit, independent of type?

peanball (dismissed):
I think the checkout with depth 0 may be counteracting the commit lint checking of a commit range. Could you please clarify?

.github/workflows/commitlint.yml
Soha-Albaghdady: The configuration enforces conventional commits : https://github.com/conventional-changelog/commitlint/tree/master/@commitlint/config-conventional#type-enum
So, only the types accepted in conventional commits will be accepted

.github/workflows/commitlint.yml
```diff
@@ -0,0 +1,31 @@
...
+      - name: Validate PR commits with commitlint
+        if: github.event_name == 'pull_request'
+        run: npx commitlint --from ${{ github.event.pull_request.head.sha }}~${{ github.event.pull_request.commits }} --to ${{ github.event.pull_request.head.sha }} --verbose
```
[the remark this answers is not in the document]
Soha-Albaghdady: depth 0 fetches all history for all tags and branches https://github.com/actions/checkout#fetch-all-history-for-all-tags-and-branches 
Why do you see this problematic?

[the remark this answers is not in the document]
peanball: I don't. just misread. `depth: 1` would have been problematic, I just mixed it up. All good.

.github/workflows/commitlint.yml
peanball: ok, but we have diverging ones in our convention. Compare also https://github.com/cloudfoundry/pcap-release/pull/81#discussion_r1259479864 (sorry, I thought I had submitted that already but it was still "pending")

peanball:
Checked with "breaking change" indicator `!` without the appropriate header in action build https://github.com/cloudfoundry/pcap-release/actions/runs/5545026407/jobs/10123338288?pr=76

works.

commitlint.config.js
```diff
@@ -0,0 +1,51 @@
+// commitlint.config.js
+module.exports = {
+    rules: {
+        // log-commit-info is used for debugging. It outputs the dict passed to a plugin.
+        // 'log-commit-info': [2, 'always'],
+        'breaking-change-indicator': [2, 'always'],
```
b1tamara, line 6: We can explain for better readability in one comment: [2 - consider as an error, 'always' - rule always applicable]

peanball: done. also linked to the documentation where this is described.

.github/workflows/commitlint.yml
```diff
@@ -0,0 +1,28 @@
...
+          fetch-depth: 0
+      - name: Setup Node.js
+        uses: actions/setup-node@v2
+        with:
+          node-version: '18.x'
+      - name: Print versions
+        run: |
+          node --version
+          npm --version
+          npx commitlint --version
```
a18e, line 22:
```suggestion
```
I don't think we need this step in a live pipeline, can be added back anytime for debugging and takes too long to execute (11s). Also: `npx commitlint --version` will always fail since it's only being installed in the next step.

.github/workflows/commitlint.yml
```diff
@@ -0,0 +1,28 @@
...
+      - name: Install commitlint
+        run: |
+          npm install -g commitlint
+      - name: Validate PR commits with @commitlint/cli
+        if: github.event_name == 'pull_request'
```
a18e, line 27: Why are we checking this at such a late point in time? (Also not sure how it works exactly/whether we actually need it?)

commitlint.config.js
```diff
@@ -0,0 +1,56 @@
+// commitlint.config.js
```
a18e, line 1:
```suggestion
```
why?

commitlint.config.js
```diff
@@ -0,0 +1,56 @@
...
+    plugins: [
+        {
+            rules: {
+                // log-commit-info is used for debugging. It outputs the dict passed to a plugin.
+                // 'log-commit-info': (input) => {
+                //     console.log(input)
+                //     return [true, "see above"]
+                // },
+                'breaking-change-indicator': ({wrong_breaking, breaking, scope, body, footer}) => {
+                    if (scope && wrong_breaking) {
```
a18e, line 39: Sounds to me like you could use `dep!!: this is still valid` because we're not checking the rule if there's no scope?

commitlint.config.js
```diff
@@ -0,0 +1,56 @@
...
+                        return [
+                            false, "The breaking change indicator `!` can only be used right before `:`, e.g. `feat(scope)!:`."
+                        ]
+                    }
+
+                    let rxBreakingChange = /^breaking(\s|-)change: .+/img
```
a18e, line 45:
```suggestion
                    const rxBreakingChange = /^breaking(\s|-)change: .+/img
```

commitlint.config.js
```diff
@@ -0,0 +1,56 @@
...
+
+                    // the marker "breaking change:" is treated as footer and "note" by commitlint, and not considered part of the body.
+                    return [
+                        !breaking || body?.match(rxBreakingChange) || footer?.match(rxBreakingChange),
```
a18e, line 49: writing this a bit more verbously wouldn't hurt but also fine as is.

a18e (changes requested):
see comments/suggestions, otherwise great!

.github/workflows/commitlint.yml
peanball: thanks. this is redundant as we're limiting this pipeline to run only on `pull_request` anyway (see above)

commitlint.config.js
peanball: Pulled out to a  variable and added explanations for the individually ORed sub-expressions

commitlint.config.js
peanball: even better: because the type was not enforced, `dep!!` was just not parsed as type and commit lint said "ok, come in."

Now I'm enforcing the type, and reworked the regex, so it will only match the indicator in the "wrong" part if there's a scope defined at all.

b1tamara (approved):
LGTM

commitlint.config.js
```diff
@@ -0,0 +1,71 @@
...
+        'subject-max-length': [2, 'always', 75],
+        // enforce the maximum scope length
+        'scope-max-length': [2, 'always', 25],
+    },
+    parserPreset: {
+        parserOpts: {
+            // add matching groups for `!` before and after scope. Before scope is invalid, after scope is correct
+            // and enforces the "breaking-change" header (see below).
+            headerPattern: /^(\w*)(?:(!?)\(([\w\$\.\-\* ]*)\))?(!?)\: (.*)$/,
+            headerCorrespondence: ["type", "wrong_breaking", "scope", "breaking", "subject"],
```
a18e, line 34:
```suggestion
            headerPattern: /^(\w*)(?:(!?)\(([\w\$\.\-\* ]*)\))?(!?)\:( ?)(.*)$/,
            headerCorrespondence: ["type", "wrong_breaking", "scope", "breaking", "subject_prefix", "subject"],
```

commitlint.config.js
```diff
@@ -0,0 +1,71 @@
...
+                // 'log-commit-info': (input) => {
+                //     console.log(input)
+                //     return [true, "see above"]
+                // },
+                'breaking-change-indicator': ({wrong_breaking, breaking, scope, body, footer}) => {
+                    if (wrong_breaking) {
+                        return [
+                            false, "The breaking change indicator `!` can only be used right before `:`, e.g. `feat(scope)!:`."
+                        ]
+                    }
```
a18e, line 50:
```suggestion
                'breaking-change-indicator': ({wrong_breaking, breaking, scope, subject_prefix, body, footer}) => {
                    if (wrong_breaking) {
                        return [
                            false, "The breaking change indicator `!` can only be used right before `:`, e.g. `feat(scope)!:`."
                        ]
                    }
                    if (!subject_prefix) {
                        return [
                            false, "A space before the commit subject is required."
                        ]
+                    }
```

a18e (approved):
If you omit the space before the subject (e.g. `dep(scope)!:Subjectstarts`), commitlint will fail with the error `type may not be empty`. 
I find this confusing/misleading. 
The following suggestions fix this but in a slightly hacky way (also the rule name `breaking-change-indicator` doesn't fit anymore).
I'm fine with the suggestion being ignored and approve of the rest of the PR.

commitlint.config.js
peanball: This is going a bit overboard for me. The rule for breaking change indicators should not be reused for another check. If we can live with a slightly confusing error message when the user is formatting the commit wrong, then we should leave it as is.

Otherwise I’d rather start a new rule checker that makes sure that all necessary segments are there. I would not want to do that though.

commitlint.config.js
peanball: > The following suggestions fix this but in a slightly hacky way (also the rule name breaking-change-indicator doesn't fit anymore).

As mentioned on the other comment, it's too hacky for me. Also for the purpose of our initial goal, i.e. "don't allow misfit commits", this does not add anything.

If you wanted, you can propose a new checker that does more than `type-empty` from the default set and inspects the commit message in full. But I see this separate to the breaking change indicator, and a non-vital improvement.

peanball:
> If you omit the space before the subject (e.g. `dep(scope)!:Subjectstarts`), commitlint will fail with the error `type may not be empty`.
> I find this confusing/misleading.
> The following suggestions fix this but in a slightly hacky way (also the rule name `breaking-change-indicator` doesn't fit anymore).
> I'm fine with the suggestion being ignored and approve of the rest of the PR.

As mentioned on the other comment, it's too hacky for me. Also for the purpose of our initial goal, i.e. "don't allow misfit commits", this does not add anything.

If you wanted, you can propose a new checker (via separate PR) that does more than type-empty from the default set and inspects the commit message and all its elements in full. But I see this separate to the breaking change indicator, and a non-vital improvement.

PR #79, opened 2023-07-10
refactor: Reimplement TLS configuration in coding and in jobs config

b1tamara:
- Instead of using one general struct for the TLS config, define ClientTLS and ServerTLS to have clear use case. 
- Adapt pcap-api spec to raise an exception if (m)TLS feature should be enabled but certificate, private key or CA is missing
- Rename property 'ca' to 'client_cas' in spec files
- Adapt acceptance tests to set common name of the Bosh Director certificate dynamically


maxmoehl (dismissed):
Approving to run the CI.

domdom82 (dismissed):

domdom82:
approved to run ci

maxmoehl (dismissed):
approving for CI

domdom82 (dismissed):

domdom82 (approved):

PR #80, opened 2023-07-10
fix(ci): use lts version of nodejs to fix docker-for-mac build

domdom82:
Fixes hanging `npm install` command.

### DIFF
--- a/ci/Dockerfile
+++ b/ci/Dockerfile
@@ -3,14 +3,15 @@ FROM bosh/docker-cpi:main
 
 ARG GINKGO_VERSION=latest
 ARG GOLANGCILINT_VERSION=latest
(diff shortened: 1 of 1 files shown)


peanball:
In a personal project I just bumped semantic-release and it insists on Node 18+ now... So we might need to be careful with bumping semantic-release.

domdom82:
@peanball LTS is still 18 so it should be OK, but you're right. I assume the tests will fail if the bump doesn't work.

peanball:
> @peanball LTS is still 18 so it should be OK, but you're right. I assume the tests will fail if the bump doesn't work.

my bad, thought it was 16. Should have checked. All good then, thanks!

peanball (dismissed):

b1tamara (dismissed):

peanball (approved):

PR #81, opened 2023-07-11
ci(test): local conventional commit check

raghunathd8:
Add a local git commit hook that checks the commit message.

Ensures that the message follows our commit convention, including breaking change indicator. This script does not have dependencies aside from basic POSIX tools.

If `commitlint` is installed, it will be used instead of the script. The configured `commitlint` rules will apply.

### DIFF
--- a/.hooks/README.md
+++ b/.hooks/README.md
@@ -1,0 +1,21 @@
+# Git Hooks
+
+Run conventional commit checks locally without any tool installation
+
+## Usage
+
+Adjust the git commit hook configuration as follows, from the root directory of pcap-release:
+
+```shell
+git config core.hooksPath .hooks
+```
+
+Please note that any hooks you had in other directories may not work.
+
+Alternatively you can also soft-link the commit-msg script:
+
+```shell
+ln -s .hooks/commit-msg .git/hooks/commit-msg
+```
+
+This invocation will fail if you have a commit-msg hook already, in order to not break existing hooks.
--- a/.hooks/commit-msg
+++ b/.hooks/commit-msg
@@ -1,0 +1,49 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+# use commitlint with its full config if available
+if command -v commitlint >/dev/null && commitlint -v >/dev/null; then
+  # Find base directory (script_location/..)
+  COMMITLINT_CONFIG="$(realpath "$(dirname "$0")/..")/commitlint.config.js"
+  # execute commitlint only if the commitlint config exists
+  [[ -f "$COMMITLINT_CONFIG" ]] && exec commitlint --edit "$1" --verbose --config "$COMMITLINT_CONFIG"
+fi
+
+# Configuration
+
+# list of allowed commit types, space separated
+TYPES="fix feat dep ci doc refactor test"
+# scope length
+SCOPE_LENGTH=25
+# commit subject length
+SUBJECT_LENGTH=70
+
+# Logic
+first_line=$(head -n1 "$1")
+conv_commit_msg="^(${TYPES// /|})(\(.{1,$SCOPE_LENGTH}\))?(!?): (.{1,$SUBJECT_LENGTH})\$"
+
+_print_convention() {
+  echo "ERROR: Commit message does not confirm to commit conventions."
+  echo "       ----"
+  sed 's/^/     > /' "$1"
+  echo
+  echo "Rules: type(scope)!: commit subject"
+  echo "     - type: must be one of: ${TYPES// /, }"
+  echo "     - scope: (optional) arbitrary text: "
+  echo "     - !: breaking change indicator. Body MUST contain 'breaking-change:' header (case in-sensitive)"
+}
+
+# trap will be fired when the script exits. It will exit if any of the commands, including the grep below, fail.
+trap '_print_convention "$1"' EXIT
+
+# check that the first line conforms to the convention pattern
+grep -qE "$conv_commit_msg" <<< "$first_line"
+
+# if there is a !, indicating breaking change, ensure that there is a "breaking-change:" header.
+if [[ "$(sed -nE "s/$conv_commit_msg/\3/p" <<<"$first_line")" == "!" ]]; then
+  grep -qiE "^breaking(-| )change: .+" "$1"
+fi
+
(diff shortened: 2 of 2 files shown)


.hooks/README.md
```diff
@@ -0,0 +1,17 @@
+# Git Hooks
+
+Run conventional commit checks locally without any tool installation
+
+## Usage
+
+Execute the below command in the root directory of the project:
+
+```sh
+ln -s -f ../../.hooks/commit-msg ./.git/hooks/commit-msg
```
maxmoehl, line 10: Where should `../..` point to? Since this moves out of the repository it seems a bit odd (and please don't put `--force` in there).

What about only configuring the local git in this repo? This can be achieved by running `git config core.hooksPath .hooks` in the root of the repo and should work like a charm. Alternatively git uses `.git/hooks` by default so you could symlink the hooks to there.

```suggestion
ln -s .hooks/commit-msg .git/hooks/commit-msg
```

```suggestion
git config core.hooksPath .hooks
```

.hooks/README.md
```diff
@@ -0,0 +1,17 @@
...
+```
+
+Make the file executable:
+
+```sh
+chmod +x ./.hooks/commit-msg
+```
```
maxmoehl, line 17: I just tested this locally and didn't have to make the file executable since it is already checked in with the executable bit set.

```suggestion
```

.hooks/commit-msg
```diff
@@ -0,0 +1,16 @@
+#!/usr/bin/env bash
+
+first_line=$(head -n1 "$1")
+echo "commit message: $first_line"
+
+conv_commit_msg="^(fix|feat|dep|ci|doc|refactor|test)(\(.{1,50}\))?: (.{1,100})\$"
```
maxmoehl, line 6: We also have to allow for the `!` before the colon in case of breaking changes (and I removed a unused capturing group):

```suggestion
conv_commit_msg="^(fix|feat|dep|ci|doc|refactor|test)(\(.{1,50}\))?!?: .{1,100}$"
```

One thing we could consider is being stricter on scope and summary, I think 150 characters in sum is too much and it shouldn't be allowed to contain all possible characters. I would propose `\w` plus ` ` and `-` (same for the summary) for allows characters.

```suggestion
conv_commit_msg="^(fix|feat|dep|ci|doc|refactor|test)(\([- \w]{1,15}\))?!?: [- \w]{5,70}$"
```

.hooks/README.md
raghunathd8: I also felt it odd from path perspective. It does not work (gives error) if I don't give the path like that. Did you try it yourself? 
`-f` option is used because `commit-msg` file does not exist in .git/hooks, otherwise one needs to `touch` the file and then link without `-f`. 
We could also use hooks path but then we don't know how individuals' `.git` is configured. In either cases we will still need to symlink. 
So considering these, I preferred to keep it simple where one needs to just run the `ln` command.

.hooks/README.md
raghunathd8: I will edit docu as `troubleshooting` for executable :)

.hooks/README.md
maxmoehl: I'm assuming that you have configured the hooks for your system centrally which is why the local ones don't work. What does the output of `git config core.hooksPath` show inside and outside the repository?

Force should not be needed if the file does not exist, according to the [man page](https://man.archlinux.org/man/ln.1#f) it is only required if you wish to overwrite the destination file.

I think the git-config approach (my second suggestion) should be the most reliable, it overwrites the global configuration of the user but that's just how these hooks work. You only have to make sure to run the command inside the repository so that it is only configured there and not somewhere else.

.hooks/commit-msg
raghunathd8: I agree to restrict usage of special characters; I will try to keep it the one you mentioned and add a few more such as `_` and `/`. I would prefer to follow some guidelines around size of optional `scope` and mandatory `subject`. I myself did not come across any such guideline. Need not be something official, but something that is widely used/accepted. Did you find or know anything in that regard?

raghunathd8: > We also have to allow for the `!` before the colon in case of breaking changes (and I removed a unused capturing group):

Yes, this is a good point, I can mandate `BREAKING CHANGE` to be added in the trailer if `!` is specified.

peanball:
Added the breaking change check and a preference for using `commitlint` and its config, if commitlint is installed and a config exists.

.hooks/commit-msg
maxmoehl: When I started our guide I didn't put any constraints on the scope because I felt like it heavily depends on the context. For example in a bosh release you will probably use different scopes compared to a pure source repository.

The subject should be a concise description of the change and the full length of the title line should be at most 70 characters, otherwise GitHub will [start wrapping it](https://github.com/therobinkim/commit-message-length/commit/788dd604d5b19b0bf9f0d8c254d4579d3b6d7b21) which hurts readability. Since the subject is a very important part of the commit message we should try to keep everything else short (hence `feat` and not `feature` as a prefix, same goes for the scope).

.hooks/README.md
```diff
@@ -0,0 +1,21 @@
...
+```shell
+git config core.hooksPath .hooks
+```
+
+Please note that any hooks you had in other directories may not work.
+
+Alternatively you can also soft-link the commit-msg script:
+
+```shell
+ln -s .hooks/commit-msg ./git/hooks/commit.msg
```
maxmoehl, line 18: Did you try this? Looks like a typo to me:
```suggestion
ln -s .hooks/commit-msg .git/hooks/commit-msg
```

And here the inverse applies: If you have hooks configured elsewhere this won't work out of the box.

peanball: seems like I haven't 😅
thanks. Will apply it as mixup.

maxmoehl (approved):

.hooks/commit-msg
peanball: I would not want to limit the type of text in scope and subject to "word characters", `-` and ` `. A bunch of other characters (;?/}{[]()*) and many more are excluded. Emoji are excluded.

Main goal is to have the release note created correctly from the commits. This is ensured with the type and breaking change indicator. The rest should be copied verbatim.

PR #83, opened 2023-07-12
fix(ci): Print pcap-api logs in case of failed specs.

domdom82:
print job logs in case of failed tests.

still a bit wip. looking to make this globally available or something.


maxmoehl:
> As this is related to ci only, could we have the commit mention `ci:` instead of `fix(ci):`?

Or, since it's changes to the test framework, it could also be `test:`.

maxmoehl (approved):

PR #86, opened 2023-07-17
ci(semantic-release): add support for our commit convention

peanball:
Based on the angular commit convention. The `angular` preset is still used, but extended / mostly overwritten for our needs.  

Special handling
 * `feat`, `fix`, `dep`, `doc`, `refactor` will show up in the release note
 * `ci`, `test` will not show up in the release note, unless they are breaking changes.
 * Any commit types that don't fit will be put in a category "Other".

Supports the breaking change indicator `!` and maps our types appropriately to sections in the release note.

## Suggestion for Testing this PR:

As the semantic-release step in dry run mode can only be run from `main` in the `rc` step of the CI pipeline, a local test is useful to check that the configuration and rules actually work as intended.

The following are instructions on how to run this new semantic-release configuration locally.

1. Comment out the following lines, as they are creating BOSH releases, etc which are not relevant for the release note:
   ```
        ["@semantic-release/exec", {
            "verifyReleaseCmd": "bosh -n create-release --final --version ${nextRelease.version} --tarball /tmp/pcap-${nextRelease.gitTag}.tgz",
            "publishCmd": "./ci/scripts/publish.sh ${nextRelease.gitTag}",
            "generateNotesCmd": "./ci/scripts/notes.sh ${nextRelease.gitTag} ${nextRelease.version}"
        }],
        ["@semantic-release/github", {
            "assets": [
                {"path": "/tmp/pcap-*.tgz"}
            ]
        }]
   ```
2. Set the "active" branch for conventional commit generation from `main` to `semantic-release-commit-convention`:
   ```diff
   - "branches": ["main", {"name": "alpha", "prerelease": true}, {"name": "beta", "prerelease": true}],
   + "branches": ["semantic-release-commit-convention", {"name": "alpha", "prerelease": true}, {"name": "beta", "prerelease": true}],
   ```
3. Add some demo commits that you want to test, e.g. with prefix `ci:`, `docs:` or a non-conventional-commit. The local commit check could be disabled / ignored, as you will not commit any of those demo commits.
4. Run semantic-release (install, if necessary):
   ```shell
   GH_TOKEN=your_github_token semantic-release --dry-run=true
   ```
   The GitHub token is needed, but in `dry-run` mode will not be used to create a release. The output of the test will be something along the lines of:
   ```
   [...]
   marked(): headerIds and headerPrefix parameters enabled by default, but are    deprecated since version 5.0.0, and will be removed in the future. To clear    this warning, install  https://www.npmjs.com/package/marked-gfm-heading-id,    or disable by setting `{headerIds: false}`.
   # 2.0.0 (2023-07-19)
   
   ### BREAKING CHANGES
   
       * well...
       * please revert this, it will break!
   
   ### Bug Fixes
   
       * *Breaking Change!* another brak! ([bfa53c2](https://github.com/cloudfoundry/pcap-release/commit/bfa53c2d37937b4e5c8027d3c239baa4e4778597))
   
   ### Documentation
   
       * add CONTRIBUTING page ([29db9c7](https://github.com/cloudfoundry/pcap-release/commit/29db9c7a6b89633f25f63926dd37034211a60c72))
   
   ### Features
   
       * *Breaking Change!* something something ([2351c67](https://github.com/cloudfoundry/pcap-release/commit/2351c671c5d1d61febfcb8c22a4e8ffcfbd267b8))
   ```


release.config.js
```diff
@@ -0,0 +1,118 @@
...
+// are valid commits and are processed correctly.
+let parserOpts = {
+    "headerPattern": /^(\w*)(?:\(([\w\$\.\-\* ]*)\))?(!?)\: (.*)$/,
+    "headerCorrespondence": ["type", "scope", "breaking", "subject"],
+    "noteKeywords": ["BREAKING CHANGE", "BREAKING-CHANGE"]
+}
+
+// commit metadata transform based on writer opts from @conventional-changelog/conventional-changelog-angular
+// https://github.com/conventional-changelog/conventional-changelog/blob/master/packages/conventional-changelog-angular/writerOpts.js
+// Adapted for our convention.
```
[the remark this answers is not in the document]
peanball: Our git page is internal, the check is open source. What I can do is to add a `CONTRIBUTING` file that explains the convention.

Good point that the convention is not public yet, thanks!

release.config.js
```diff
@@ -0,0 +1,118 @@
...
+        note.title = 'BREAKING CHANGES'
+        discard = false
+    })
+
+    if (commit.type === 'feat') {
+        commit.type = 'Features'
+    } else if (commit.type === 'fix') {
+        commit.type = 'Bug Fixes'
+    } else if (discard) {
+        return
```
Soha-Albaghdady, line 27: As discussed, instead of discarding commits that don't match the conventional commit, lets introduce an "Other" section and add them there. We can remove the section when we are sure that it is always empty (i.e. commitlint is doing its intended job)

Soha-Albaghdady (changes requested):

release.config.js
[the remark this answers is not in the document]
Soha-Albaghdady: Ah yes, you are right. We should publish the convention somewhere. +1 for the `CONTRIBUTING` file, we do need it anyways

release.config.js
peanball: And even better, when there’s no need for a section, it just doesn’t show up.

The fallback would still be there.

release.config.js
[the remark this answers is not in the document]
peanball: Added a `CONTRIBUTING` file.

CONTRIBUTING
```diff
@@ -0,0 +1,80 @@
...
+
+co-authored-by: John Doe <john.doe@example.com>
+breaking change: this is a major change that leads to a new version
+```
+
+Many of the elements are optional, but type and subject are mandatory.
+
+The following elements exist and can be used:
+* `type`: a classification of the change, e.g. whether it's a new feature, bug fix or behind-the-scenes change. See below.
+* `subject`: a sub-classification of the change, often used to describe the component affected by a change.
```
Soha-Albaghdady, line 26: This should be 'scope' not subject.

Soha-Albaghdady (changes requested):

release.config.js
maxmoehl: I would focus on the types of changes that matter to our consumers:
- all commits that change the artefacts we are shipping, since those are potential sources of bugs / behavioural changes (`feat`, `fix`, `refactor`, `dep`)
- documentation changes (`doc`); with a bosh release these are also part of the build output, so they also qualify for the first bullet point

`test` and `ci` do not (or at least should not) affect the build output and therefore shouldn't affect our consumers. As such I don't see the need to include them in the release notes.

CONTRIBUTING
peanball: thanks!

peanball: addressed.

release.config.js
peanball: I fully agree with that. There are now two separate lists of commit type to heading mapping, one for "changes" that always show up in the release note, and one for "hidden" commits that only show up if they have a breaking change.

Mapping is exactly as you laid out.

Any commits not fitting into that schema will show up in "Other", as discussed.

a18e:
I think it would make sense to have the `### BREAKING CHANGES` section at the top of the release note. It's the most important section after all and might get overlooked otherwise.

peanball:
> I think it would make sense to have the `### BREAKING CHANGES` section at the top of the release note. It's the most important section after all and might get overlooked otherwise.

Ok, that works out fine. I am considering also adding an indicator for the commits themselves, like this:

# 2.0.0 (2023-07-19)

### BREAKING CHANGES

    * well...
    * please revert this, it will break!

### Bug Fixes

    * *Breaking Change!* another brak! ([bfa53c2](https://github.com/cloudfoundry/pcap-release/commit/bfa53c2d37937b4e5c8027d3c239baa4e4778597))

### Documentation

    * add CONTRIBUTING page ([29db9c7](https://github.com/cloudfoundry/pcap-release/commit/29db9c7a6b89633f25f63926dd37034211a60c72))

### Features

    * *Breaking Change!* something something ([2351c67](https://github.com/cloudfoundry/pcap-release/commit/2351c671c5d1d61febfcb8c22a4e8ffcfbd267b8))

domdom82 (changes requested):

CONTRIBUTING
```diff
@@ -0,0 +1,80 @@
...
+
+A conventional commit is structured as follows:
+
+```plain
+type(scope)!: add an imperative subject with effects of the change
+
+Other text in the body describes the change
+
+co-authored-by: John Doe <john.doe@example.com>
+breaking change: this is a major change that leads to a new version
```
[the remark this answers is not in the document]
peanball: semantic-release calls this a "Note", so does their angular convention parser.

Technically us using this _as a footer_ is the wrong bit. I can change the example though.

CONTRIBUTING
```diff
@@ -0,0 +1,80 @@
...
+The working group for application runtime, and the network extensions team may approve PRs.
+
+## Continuous Integration
+
+Continuous integration is run at https://concourse.cfi.sapcloud.io. 
+
+
+
+
+We are working on running acceptance tests without approved PRs
```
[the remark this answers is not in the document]
peanball: thanks. this line is an accidental leftover anyway.

domdom82 (approved):

PR #87, opened 2023-07-18
fix: Make sure pcap is deployable via manifest and ops files.

domdom82:
- make pcap-agent an addon
- remove unused cert / ca variables from pcap-agent ops file (we are using the ca from pcap-api.yml)
- backport fixes to pcap-api.yml
- remove unused pcap_api_bosh_mtls cert
- update vars-template so that it actually works
- update README.me to reflect the current state

### DIFF
--- a/manifests/pcap-api.yml
+++ b/manifests/pcap-api.yml
@@ -7,6 +7,9 @@ addons:
       - name: bpm
         release: bpm
   - name: bosh-dns-aliases
+    include:
+      instance_groups:
+        - pcap-api
     jobs:
       - name: bosh-dns-aliases
         release: bosh-dns-aliases
@@ -39,7 +42,7 @@ instance_groups:
         release: pcap
         properties:
           pcap-api:
-            log_level: info
+            log_level: debug
             buffer:
               size: 1000
               upper_limit: 995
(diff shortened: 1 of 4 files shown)


manifests/pcap-api.yml
```diff
@@ -48,19 +51,23 @@ instance_groups:
...
                 skip_verify: false
-                ca: ((/bootstrap-bosh/bosh/default_ca.ca))
+                ca: ((pcap_api.bosh_director_ca))
             agents_mtls:
               common_name: pcap-agent.service.cf.internal
               skip_verify: false
-              certificate: ((pcap_api_mtls.certificate))
+              certificate: ((pcap_api_mtls.certificate))((pcap_agent_mtls.ca))
               private_key: ((pcap_api_mtls.private_key))
               ca: ((pcap_agent_mtls.ca))
```
b1tamara, line 72: We can use here the vars from vars-template.yml to be consistent.

domdom82: what do you mean? the `pcap_api_mtls` certificate has no definition in the vars-template.yml file.
The other entries are there only because they are not defined in the manifest and have to be imported from other deployments such as `cf`.

b1tamara: you are right.

b1tamara (approved):
LGTM

PR #105, opened 2023-08-04
refactor(api): Make maxConcurrentCaptures int32 to avoid cast.

domdom82:
Minor refactor. We compare maxConcurrentCaptures to concurrentStreams which should match datatypes.

### DIFF
--- a/src/pcap/cmd/pcap-api/config.go
+++ b/src/pcap/cmd/pcap-api/config.go
@@ -30,7 +30,7 @@ var DefaultAPIConfig = APIConfig{
 type APIConfig struct {
 	pcap.NodeConfig    `yaml:"-,inline"`
 	AgentsMTLS         *pcap.ClientTLS `yaml:"agents_mtls" validate:"omitempty"`
-	ConcurrentCaptures uint            `yaml:"concurrent_captures"`
+	ConcurrentCaptures int32           `yaml:"concurrent_captures"`
 	DrainTimeout       time.Duration   `yaml:"drain_timeout"`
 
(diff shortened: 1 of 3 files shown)


maxmoehl (commented):
I'd prefer `unit32` since the value can't be negative.

src/pcap/cmd/pcap-api/config.go
```diff
@@ -30,7 +30,7 @@ var DefaultAPIConfig = APIConfig{
 type APIConfig struct {
 	pcap.NodeConfig    `yaml:"-,inline"`
 	AgentsMTLS         *pcap.ClientTLS `yaml:"agents_mtls" validate:"omitempty"`
-	ConcurrentCaptures uint            `yaml:"concurrent_captures"`
+	ConcurrentCaptures int32           `yaml:"concurrent_captures"`
```
maxmoehl, line 33: Missed one :)

domdom82: yep.

maxmoehl:
> > I'd prefer `unit32` since the value can't be negative.
> 
> do you mean uint32? it's possible (as I've shown in the latest commit), however it's non-trivial to subtract from an unsigned int. See my comment in [api.go](https://github.com/cloudfoundry/pcap-release/pull/105/files#diff-a0a2961061d946ca261bdfcdf90fb7c886ff6e9f0d04c43c8c0dfc4812e9a2f9R183). You can't add uint32(-1) so you have to use the binary complement to do that. Not sure if it's worth the confusion this will likely cause...

Well in that case we can probably accept a `int32` as that makes the code a lot more readable 😄

maxmoehl (approved):

PR #106, opened 2023-08-04
feat: Add support for local development with TLS enabled.

domdom82:
This PR adds the option to run all pcap components locally with TLS enabled.
Two new run configurations:
- [.run/mock-bosh-setup-tls.run.xml](https://github.com/cloudfoundry/pcap-release/pull/106/files#diff-b166b6631b1403f4d0e7042b1632d97abb35b2c68d4f146461c19fc68fe7d805) This runs the mock BOSH director and generates pcap-api and pcap-agent configs with TLS enabled. It also generates all certificates needed.
- [.run/bosh-cli_local-tls.run.xml](https://github.com/cloudfoundry/pcap-release/pull/106/files#diff-3797f2305f617520682a8a80e3c8c7e697cd681748f4e128ae0b1aa3011b8dba) This starts the CLI with the `insecure` flag which is necessary because the pcap-api certificate is self-signed.

Also a `ca.go` file is added to take care of making certificates for the local TLS case. I felt like using step here would be overkill for just the local dev case. If we find we need to support more complex PKIs, using [step certificates](https://github.com/smallstep/certificates) may be viable.

### DIFF
--- a/src/pcap/devtools/mock-bosh-setup/ca.go
+++ b/src/pcap/devtools/mock-bosh-setup/ca.go
@@ -1,0 +1,209 @@
+package main
+
+import (
+	"bytes"
+	"crypto/rand"
+	"crypto/rsa"
+	"crypto/x509"
+	"crypto/x509/pkix"
+	"encoding/pem"
+	"io"
+	"math/big"
+	"os"
+	"time"
+)
+
+// Some helper tools to generate x509 CAs and Certificates
+
+const RootCABits = 4096
+const ServerCertBits = 2048
+const ClientCertBits = 2048
+const RootCASerial = 1337
+const ServerCertSerial = 3141
+const ClientCertSerial = 2718
+
+type CA struct {
+	Cert    *x509.Certificate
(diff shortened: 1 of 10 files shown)


src/pcap/devtools/mock-bosh-setup/ca.go
```diff
@@ -0,0 +1,202 @@
...
+	"crypto/x509"
+	"crypto/x509/pkix"
+	"encoding/pem"
+	"io"
+	"math/big"
+	"os"
+	"time"
+)
+
+// Some helper tools to generate x509 CAs and Certificates
```
peanball, line 16: This is not a lot of documentation for this much code.

A few godoc comments describing what the specific stuct or function do would be great. A lot of it is self-explaining, fair enough.

For the certificate creation (newCA, ClientCert), some of the key points would be helpful, e.g. "uses the provided CA to create a new client certificate. The appropriate usage flags for client authentication and signature are set."

peanball (changes requested):

peanball (approved):

PR #111, opened 2023-08-09
fix: CLI not stopping on CTRL+C

maxmoehl:
This is a hack to showcase the behaviour that we need in order to resolve the bug. A proper implementation will follow.


plowin:
continued via https://github.com/cloudfoundry/pcap-release/pull/113

PR #113, opened 2023-08-16
fix: bosh CLI not stopping on CTRL+C

b1tamara:


src/pcap/client.go
```diff
@@ -172,7 +169,8 @@ func (c *Client) ProcessCapture(ctx context.Context, endpointRequest *EndpointRe
 		},
 	}
 
-	c.stream, err = c.Capture(ctx)
+	apiContext := context.Background()
+	c.stream, err = c.Capture(apiContext)
```
maxmoehl, line 173:
```suggestion
	c.stream, err = c.Capture(context.Background())
```

maxmoehl (changes requested):
You mainly aligned the client side implementation with the API->Agent logic, right?

Looks good overall, is there a way we can add a test case for this to catch changes that would break this logic again?

src/pcap/client.go
```diff
@@ -244,18 +247,20 @@ func (c *Client) ReadCaptureResponse(stream API_CaptureClient, packetWriter *pca
...
 			res, err := stream.Recv()
 			if errors.Is(err, io.EOF) {
 				logger.Info("clean stop, done")
 				cancel(nil)
-				break
+				return
 			}
+
 			code := status.Code(err)
 			if code != codes.OK {
```
peanball, line 260: Do we need to check for an `OK` status as well?

src/pcap/client.go
```diff
@@ -233,6 +234,8 @@ func (c *Client) StopRequest() {
 	if err != nil {
 		c.log.Error("could not stop")
 	}
+	c.stream.CloseSend()
```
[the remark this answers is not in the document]
b1tamara: We have a bidi stream. CloseSend closes the send direction (from client to server) of the stream to tell the server that we won’t send any more data.

src/pcap/client.go
b1tamara: @peanball Why do you think we need the status OK? We are checking here for on err from stream.Recv but if there are no errors, we only read the packet/msg

b1tamara:
@maxmoehl Yes. Now the implementation of client is very similar to api and agent (in functionality of main _capture_ function and in reading from stream)

maxmoehl (approved):

PR #168, opened 2023-11-13
ci: limit number of open PRs

b1tamara:


.github/dependabot.yml
```diff
@@ -15,9 +17,11 @@ updates:
   directory: /src/pcap
   schedule:
     interval: weekly
```
maxmoehl, line 19: If we now limit this to one open PR, we should decrease the interval:
```suggestion
    interval: daily
```
Otherwise we have to wait a full week for the next PR and risk piling up dependency updates.

Applies to the go updates below as well.

maxmoehl (changes requested):

maxmoehl (approved):

.github/dependabot.yml
b1tamara: applied

PR #187, opened 2024-02-05
build with vendor directory

fredga:
-  In a production environnent internet access is not available.
Building the pcap-agent and pcap-api packages is not possible because the dependencies are taken from the internet.

- This PR consists of using the go vendor directory in order to vendorize the dependencies before the build phase of these packages.


maxmoehl:
Thank you for your contribution!

Have you tested this change? I would expect that you also have to check-in the vendor directory to git, otherwise the dependencies won't be available when running the packaging script.

fredga:
Hi,
I made a new commit with the deletion of the vendor in .gitignore file and i have test the compilation step and it is correct.

maxmoehl:
Thanks, that looks good now! Could you please squash your changes into a single commit according to our [CONTRIBUTING](https://github.com/cloudfoundry/pcap-release/blob/main/CONTRIBUTING) guide? This is needed for our automated release process, if you are unsure what to put I can also provide you some examples.

plowin:
Hi @fredga, if this is still relevant, please resolve the conflicts and apply review feedback. Or give us a short comment if and how you worked around it.

plowin:
CF environments we are aware of can access the internet. As the described use-case seems to be rare and the author worked around the issue, we are closing the PR. Feel free to re-open and clean it up if this is still required. Please also add some more details on your setup without internet access.

PR #197, opened 2024-02-28
ci: create github workflow to auto-merge dependency PRs

b1tamara:


maxmoehl:
Relevant documentation: https://docs.github.com/en/code-security/dependabot/working-with-dependabot/automating-dependabot-with-github-actions#approve-a-pull-request

maxmoehl (dismissed):

.github/workflows/automerge.yml
```diff
@@ -0,0 +1,22 @@
+name: Auto-merge dependency PRs
+on: pull_request
```
maxmoehl, line 2: Wouldn't [`pull_request_target`](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request_target) be the safer option in this context?

maxmoehl: If my understanding is correct, a PR could modify this workflow to run for any PR and get an approval when this is using `pull_request`.

peanball: it will not work for PRs from a fork.

Also `pull_request` does not imply any permissions, while `pull_request_target` implies read/write to the PR. We were using `pull_request_target` on the sample apps specifically because it actually can access secrets and gains more permissions, and comes with caveats.

So if `pull_request` works and provides enough permissions to do the intended task, I would stick with that.

b1tamara: Yes I explicitly decided for `pull_request` as it's fully enough for our purpose:

pull_request event triggers the workflow when activity related to Pull Requests happen. However, if the pull request is from a fork, the workflow will run with a read-only permission. Also, secrets are not passed to the runner to avoid exposing sensitive data.

maxmoehl: Thanks, that makes sense!

maxmoehl (approved):

PR #266, opened 2024-11-28
Fix ci build

mtekel:
Using updated pcap-testflight image, bump ruby spec to match image version. Bump acceptance tests golang. Address issues highlighted by new lint version.

Changes can be viewed commit-by-commit for more focused/limited scope of modifications.


src/pcap/test/bosh_test.go
```diff
@@ -74,7 +74,7 @@ func TestNewBoshResolver(t *testing.T) {
 func TestAuthenticate(t *testing.T) {
 	bar, _, _, err := mock.NewResolverWithMockBoshAPI(nil)
 	if err != nil {
-		t.Errorf(err.Error())
+		t.Errorf("%s", err.Error())
```
peanball, line 77:
```suggestion
		t.Error(err)
```

src/pcap/api_test.go
```diff
@@ -306,7 +306,7 @@ func TestCapture(t *testing.T) {
 				t.Errorf("capture() unexpected error during api creation: %v", err)
 			}
 
-			var connectToTargetFn = func(ctx context.Context, req *CaptureOptions, target AgentEndpoint, creds credentials.TransportCredentials, log *zap.Logger) (captureStream, error) {
+			var connectToTargetFn = func(ctx context.Context, req *CaptureOptions, target AgentEndpoint, creds credentials.TransportCredentials, log *zap.Logger) (captureStream, error) { //nolint:revive //keep vars even if unused for better context
 				return tt.stream, tt.err
```
peanball, line 310: I would argue based on the type you could still figure out what is put there. The name is not strictly needed.

src/pcap/test/integration/helpers.go
```diff
@@ -176,17 +176,17 @@ func generateCerts(commonName string, dir string) (string, string, string, error
 	certPath := path.Join(dir, "cert.pem")
 	keyPath := path.Join(dir, "private.key")
 	caPath := path.Join(dir, "ca.pem")
-	err = os.WriteFile(certPath, certPEM.Bytes(), os.ModePerm)
+	err = os.WriteFile(certPath, certPEM.Bytes(), 0600)
 	if err != nil {
```
peanball, line 180: I would prefer defining a variable with a name that describes the permissions, then use that. A 'magic number' linter should have noted this.

src/pcap/agent.go
```diff
@@ -115,7 +115,7 @@ func (a *Agent) Capture(stream Agent_CaptureServer) (err error) {
 		return errorf(codes.InvalidArgument, "%w", err)
 	}
 
-	opts := req.Payload.(*AgentRequest_Start).Start.Capture
+	opts := req.Payload.(*AgentRequest_Start).Start.Capture //nolint:errcheck //this only returns one value
```
peanball, line 118: we could have a check here for the cast.
```suggestion
	var opts *CaptureOptions
	if start, ok := req.Payload.(*AgentRequest_Start); ok {
		opts = start.Start.Capture
	} else {
		return errorf(codes.InvalidArgument, "invalid message: expected Payload of type StartAgentCapture")
	}
```
(not tested)

peanball (changes requested):

peanball:
some minor suggestions

src/pcap/api_test.go
mtekel: For some reason original author decided to do it this way, so I'll leave it be.

src/pcap/agent.go
peanball: Ok. Maybe we can eventually return the cast result, but not in this small linter fix

peanball (approved):
looks good overall. approving for CI and we can merge if all of it passes :-)

